PR #1, opened 2025-05-31
Add Dockerfile and CI for hermetic builds

moul:
Adds Dockerfile and GitHub Actions workflow to verify hermetic builds. The setup includes:
- Multi-stage Dockerfile with test stage
- CI workflow that builds and runs tests


cerisier:
Closing this since the full hermeticity is now done via remote-exec workers inside docker already using buildbuddy.

PR #44, opened 2025-11-02
Implement ubsan

### LINKED ISSUE #11: Compile and link against UndefinedBehaviorSanitizer (UBSan)
Detects various types of undefined behavior such as null pointer dereferencing, signed integer overflow, division by zero, invalid shifts, and misaligned pointer access.

Flag: -fsanitize=undefined
Lib: libclang_rt.ubsan

dzbarsky:
It's a bit tricky because we want to compile stage2 libs with sanitizers, but need to take a bit of care to avoid cycles. 

Fixes #11

### DIFF
--- a/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
+++ b/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
@@ -486,182 +486,6 @@ cc_stage2_static_library(
 
 ##### Sanitizers #######
 
-# TODO(zbarsky): It would be nice to not have to jam everything into a single BUILD file
-
-SANITIZER_COMMON_SRCS = [
-    "sancov_flags.cpp",
-    "sancov_flags.h",
-    "sanitizer_addrhashmap.h",
-    "sanitizer_allocator.cpp",
-    "sanitizer_allocator.h",
-    "sanitizer_allocator_checks.cpp",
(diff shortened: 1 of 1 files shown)


MODULE.bazel
```diff
@@ -113,12 +113,12 @@ archive_override(
     strip_prefix = "libstdcxx-stubs-generator-{}".format(LIBSTDCXX_STUBS_GENERATOR_COMMIT),
 )
 
-bazel_dep(name = "rules_python", version = "0.40.0", dev_dependency = True)
+bazel_dep(name = "rules_python", version = "0.40.0")
```
dzbarsky, line 116: there are all no-longer dev deps as we need `"@llvm-project//llvm:Symbolize"` as a dependency in the sanitizers

toolchain/stage2/args/BUILD.bazel
```diff
@@ -1,4 +1,7 @@
 load("@rules_cc//cc/toolchains:args.bzl", "cc_args")
+load("@rules_cc//cc/toolchains/impl:documented_api.bzl", "cc_args_list")
+
+package(default_visibility = ["//toolchain/stage2:__subpackages__"])
```
dzbarsky, line 4: prevent using these in the wrong spot by accident

PR #52, opened 2025-11-15
Add a few flags for determinism + reduced spew

dzbarsky:
Cribbed from https://github.com/modular/modular/blob/4458a95ede5ea3fae2043b6b260c4c55dbc798bb/bazel/internal/cc-toolchain/args/BUILD.bazel#L109 and https://github.com/modular/modular/blob/4458a95ede5ea3fae2043b6b260c4c55dbc798bb/bazel/internal/cc-toolchain/args/BUILD.bazel#L183-L184

### DIFF
--- a/toolchain/args/macos/BUILD.bazel
+++ b/toolchain/args/macos/BUILD.bazel
@@ -31,6 +31,8 @@ cc_args(
     ],
     args = [
         "-headerpad_max_install_names",
+        "-Wl,-no_warn_duplicate_libraries",
+        "-Wl,-oso_prefix,.",  # Strip absolute paths from debug info file references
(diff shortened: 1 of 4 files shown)


toolchain/args/macos/BUILD.bazel
```diff
@@ -31,6 +31,8 @@ cc_args(
     ],
     args = [
         "-headerpad_max_install_names",
+        "-Wl,-no_warn_duplicate_libraries",
+        "-Wl,-oso_prefix,.",  # Strip absolute paths from debug info file references
```
cerisier, line 35: I was thinking of that yesterday ah !

PR #53, opened 2025-11-17
Provide static_runtime_lib/dynamic_runtime_lib to toolchain

dzbarsky:
The TL;DR is this lets bazel decide whether to link static or dynamic libc++, libcc++abi, and libunwind, which come from the cc_toolchain. And then we provide empty stub `libc++.a`/etc so that `-lc++` from user-provided flags doesn't error out, though it also has no effect.

This unblocks properly supporting --dynamic_mode flag and scenarios where you link some dylib that dynamically links libc++ so you also want to dynamically link, etc

### DIFF
--- a/toolchain/args/linux/BUILD.bazel
+++ b/toolchain/args/linux/BUILD.bazel
@@ -136,8 +136,6 @@ cc_args(
         # In prevention of when all of those can be shared libraries.
         "-Wl,--push-state",
         "-Wl,--as-needed",
-        "-lc++",
-        "-lc++abi",
 
         # In newer versions of glibc various libs such as libm, librt are
         # normally included in libc and not available as separate libraries.
@@ -157,7 +155,6 @@ cc_args(
         "-lrt",
         "-lutil",
 
-        "-lunwind", # only if libcxx on linux
         "-Wl,--pop-state",
     ],
     data = [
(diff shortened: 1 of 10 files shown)


toolchain/args/linux/BUILD.bazel
```diff
@@ -131,13 +131,9 @@ cc_args(
         "-L{libunwind_library_search_path}",
         "-L.", # needed to -l:bazel-out/...
 
-        "{clang_rt.builtins.a}",
-
         # In prevention of when all of those can be shared libraries.
         "-Wl,--push-state",
         "-Wl,--as-needed",
-        "-lc++",
-        "-lc++abi",
```
cerisier, line 140: Does it still work for downstream to use -lc++ ? I ask this because a lot projects, either rust or using rules_foreign_cc do pass arbitrary -l flags for stuff they just expect to be there and we need to support this.

toolchain/args/linux/BUILD.bazel
```diff
@@ -157,18 +153,15 @@ cc_args(
         "-lrt",
         "-lutil",
 
-        "-lunwind", # only if libcxx on linux
```
cerisier, line 160: same comment

toolchain/args/linux/BUILD.bazel
cerisier: It seems the library search directories are still there so i guess it works ? I'd love to find a target that does this or maybe make one as part of the test suite ?

dzbarsky: turned these all into stubs

PR #55, opened 2025-11-18
WIP: ThinLTO feature

dzbarsky:
`bazel build //tests:for_linux_amd64_musl '--config=bootstrap'`


patches/rules_cc_preprocessor.patch
```diff
@@ -0,0 +1,143 @@
...
++    actions = None,
+     type = types.list(types.directory),
+ )
+ 
+@@ -423,7 +423,7 @@
+ 
+ cc_variable(
+     name = "system_include_paths",
+-    actions = ["//cc/toolchains/actions:compile_actions"],
++    actions = None,
```
cerisier, line 140: So we will never get any bazel provided system_include_paths ?

third_party/llvm-project/20.x/libcxx/BUILD.tpl
```diff
@@ -1395,11 +1396,11 @@ cc_stage2_static_library(
...
 
-cc_stage2_shared_library(
-    name = "libcxx.shared",
-    deps = [
-        ":libcxx",
-    ],
-    shared_lib_name = "libc++.so.1.0",
-    visibility = ["//visibility:public"],
-)
+#cc_stage2_shared_library(
```
cerisier, line 1399: removed for now ?

toolchain/args/linux/BUILD.bazel
```diff
@@ -224,45 +217,35 @@ cc_args(
...
-        "-lc",
-        "-ldl",
-        "-lrt",
-        "-lutil",
+        #"-lm",
+        #"-lpthread",
+        #"-lc",
+        #"-ldl",
+        #"-lrt",
+        #"-lutil",
```
cerisier, line 239: Same comment as #53 

Those are expected to be existing and working by a lot of downstream targets.

patches/rules_cc_preprocessor.patch
dzbarsky: I suspect this may actually mean all actions. I think this was Codex trying to get shit to compile :) will clean things up as needed once we get it actually working

third_party/llvm-project/20.x/libcxx/BUILD.tpl
dzbarsky: Yeah I saw duplicate actions created between the libs, can come back to this when cleaning up

Issue #45, opened 2025-11-04
Support for rules_go


sluongng:
As mentioned in https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/41, I had a difficult time invoking this cc toolchain through the Go linker.

There seems to be a "duplicate symbol" issue wrt which file was included at link time.

I think it would be best if the repo provides a rules_go example and runs tests (through BazelCI or through BuildBuddy Workflows) regularly against such a setup to prevent regression.

cerisier:
FYI someone reported a different issue, likely from the same root cause:

```
INFO: Analyzed 2 targets (0 packages loaded, 0 targets configured).
SUBCOMMAND: # //main:main [action 'GoLink main/main_/main', configuration: 7f8856bb0405540c89cb61a09c1d0562ac1a95a54b0f8a7e6cabe690edf58098, execution platform: @@platforms//host:host, mnemonic: GoLink]
(cd /private/var/tmp/_bazel_corentinkerisit/d376e2bcebaa8121c841204a3d92fd4a/execroot/_main && \
  exec env - \
    CGO_ENABLED=1 \
    GOARCH=arm64 \
    GODEBUG='winsymlink=0' \
    GOEXPERIMENT='' \
    GOOS=darwin \
    GOPATH='' \
    GOROOT=bazel-out/darwin_arm64-fastbuild/bin/external/rules_go+/stdlib_ \
    GOROOT_FINAL=GOROOT \
    GOTOOLCHAIN=local \
    PATH=external/toolchains_llvm_bootstrapped++_repo_rules+llvm-toolchain-minimal-21.1.4-darwin-arm64/bin:/bin:/usr/bin \
    ZERO_AR_DATE=1 \
[...]

PR #65, opened 2025-11-26
Support rules_go

cerisier:
Base for implementation of #45 

The first issue we had was linked to duplicated linker args produced by the go linker driver:

```
ld.lld: error: duplicate symbol: __dso_handle
>>> defined at crtbegin.c
>>>            crtbegin.o:(__dso_handle) in archive bazel-out/linux_x86_64-fastbuild-ST-ca5ab100789a/bin/external/toolchains_llvm_bootstrapped++http_archive+compiler-rt/clang_rt.crtbegin.static_/libclang_rt.crtbegin.static.a
>>> defined at crtbegin.c
>>>            crtbegin.o:(.data+0x0) in archive bazel-out/linux_x86_64-fastbuild-ST-ca5ab100789a/bin/external/toolchains_llvm_bootstrapped++http_archive+compiler-rt/clang_rt.crtbegin.static_/libclang_rt.crtbegin.static.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

because CRTs are being passed multiple times by `go tool link`.

The next issue is described below...


examples/rules_go/BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+    visibility = ["//visibility:public"],
+)
+
+go_cross_binary(
+    name = "cgo_linux_amd64",
+    target = "cgo",
+    platform = "@toolchains_llvm_bootstrapped//platforms:linux_x86_64",
+)
+
+build_test(
```
dzbarsky: nit: you don't really need `build_test`, `bazel test` will already ensure it builds. The only edge case is if you have no test targets, then `bazel test` will fail even if they all compile. But we do have a `go_test` here, so we're good

examples/rules_go/MODULE.bazel
```diff
@@ -0,0 +1,21 @@
...
+    name = "hermetic_cc_toolchain_bcr_tests",
+    version = "0.1.0",
+)
+
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "rules_go", version = "0.59.0")
+bazel_dep(name = "bazel_skylib", version = "1.8.2")
+
+# https://github.com/cerisier/toolchains_llvm_bootstrapped/releases
+bazel_dep(name = "toolchains_llvm_bootstrapped", version = "0.0.0", dev_dependency = True)
```
dzbarsky: you can snip `dev_dependency` here and below, it doesn't really matter :)

examples/rules_go/BUILD.bazel
```diff
@@ -0,0 +1,41 @@
...
+    srcs = ["cgo_test.go"],
+    embed = [":cgo_lib"],
+)
+
+go_binary(
+    name = "cgo",
+    embed = [":cgo_lib"],
+    gc_linkopts = select({
+        "@platforms//os:macos": [
+            "-w",  # https://github.com/ziglang/zig/issues/15439
```
dzbarsky: are these zig issues relevant?

examples/rules_go/BUILD.bazel
cerisier: Even if'm testing the cross compiled target ? the go_test seems to only test for exec.

e2e/rules_go/cgo.go
```diff
@@ -0,0 +1,14 @@
+package main
```
dzbarsky, line 1: (above) remove `a.out` from git

dzbarsky (approved):

examples/rules_go/BUILD.bazel
cerisier: Not for us, removed.

examples/rules_go/BUILD.bazel
dzbarsky: yeah if you do `bazel test //...` in this module it will still build the binaries. You can double-check by adding a syntax error and you'll see it fails, even if the `go_test` target is commented out.

not a huge deal either way though!

PR #72, opened 2025-11-30
Implement CI

cerisier:


.github/workflows/ci.bazelrc
```diff
@@ -0,0 +1,19 @@
...
+common --config remote
+
+# This directory is configured in GitHub actions to be persisted between runs.
+# docs: https://bazel.build/reference/command-line-reference#flag--disk_cache
+# build --disk_cache=~/.cache/bazel
+# docs: https://bazel.build/reference/command-line-reference#flag--repository_cache
+build --repository_cache=~/.cache/bazel-repo
+
+# docs: https://bazel.build/reference/command-line-reference#flag--remote_download_outputs
+build --remote_download_outputs=toplevel
```
dzbarsky: set [--remote_download_minimal](https://bazel.build/reference/command-line-reference#common_options-flag--remote_download_minimal) instead. We don't need to download anything given the build is fully remote

also swap these to `common` flags instead of `build`

MODULE.bazel
```diff
@@ -5,6 +5,9 @@ module(
     compatibility_level = 0,
 )
 
+bazel_dep(name = "buildifier_prebuilt", version = "8.2.1", dev_dependency = True)
```
dzbarsky: uber-nit: maybe put dev_deps at the bottom. or i'm just OCD :)

e2e/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+)
+
+#TODO(cerisier): should the toolchain provide its own list ?
+PLATFORMS = [
+    "@toolchains_llvm_bootstrapped//platforms:linux_x86_64",
+    "@toolchains_llvm_bootstrapped//platforms:linux_aarch64",
+    # "@toolchains_llvm_bootstrapped//platforms:macos_x86_64",
+    # "@toolchains_llvm_bootstrapped//platforms:macos_aarch64",
+] + [
+    "@toolchains_llvm_bootstrapped//platforms/libc_aware:linux_x86_64_{}".format(libc) for libc in LIBCS
```
dzbarsky, line 27: should we add musl to here too?

e2e/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_cc//cc:cc_library.bzl", "cc_library")
+load("@toolchains_llvm_bootstrapped//constraints/libc:libc_versions.bzl", "LIBCS")
+load("@bazel_lib//lib:transitions.bzl", "platform_transition_binary")
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+
+cc_library(
```
dzbarsky, line 7: there's also unwind.cc which appears unused?

e2e/cross_compilation/BUILD.bazel
```diff
@@ -0,0 +1,46 @@
...
+    # "@toolchains_llvm_bootstrapped//platforms:macos_x86_64",
+    # "@toolchains_llvm_bootstrapped//platforms:macos_aarch64",
+] + [
+    "@toolchains_llvm_bootstrapped//platforms/libc_aware:linux_x86_64_{}".format(libc) for libc in LIBCS
+] + [
+    "@toolchains_llvm_bootstrapped//platforms/libc_aware:linux_aarch64_{}".format(libc) for libc in LIBCS
+]
+
+[
+    platform_transition_binary(
```
dzbarsky, line 33: maybe we should have a macro here as well like `rust_binary_cross_build_test_suite`? Especially if we add more binaries like unwind.cc?

e2e/rules_rust/rust_binary_cross_build_test_suite.bzl
```diff
@@ -2,11 +2,21 @@ load("@rules_rust//rust:defs.bzl", "rust_binary")
...
+    platform = kwargs.get("platform", None)
     rust_binary(
         name = name,
-        **kwargs,
+        **kwargs
     )
-    platform = kwargs.get("platform", None)
+
+    # Temporary hack to get an exec compatible binary to run inside a sh_test.
+    native.genrule(
```
dzbarsky: nit: you can just stick this in the BUILD file, it doesn't need to be per-macro-invocation

tools/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
+load("@bazelrc-preset.bzl", "bazelrc_preset")
+load("@buildifier_prebuilt//:rules.bzl", "buildifier", "buildifier_test")
+
+buildifier(
```
dzbarsky, line 4: can we add `tools` to export-ignore like https://github.com/dzbarsky/rules_rs/blob/main/.gitattributes#L4 ?
that way we don't end up with unbuildable targets in the module

e2e/cross_compilation/BUILD.bazel
cerisier: musl is included in LIBCS already 😁

MODULE.bazel
cerisier: haha sure

tools/BUILD.bazel
cerisier: Sure

.github/workflows/ci.bazelrc
cerisier: done

e2e/cross_compilation/BUILD.bazel
cerisier: I'll merge now as discussed and we we wil look at this when we introduce other tests.

PR #73, opened 2025-11-30
Add wasm target

dzbarsky:
Implements #25


toolchain/llvm/BUILD.llvm_release.tpl
```diff
@@ -102,6 +103,7 @@ exports_files([
     "bin/ld.lld",
     "bin/ld64.lld",
     "bin/wasm-ld",
+    "bin/wasm-ld",
```
cerisier: wasm-ld was already there it seems

dzbarsky: ummm I swear it wasn't there last night when I added it, so confusing. But that's great, it means we don't need to split this into multiple changes!

cerisier (approved):
Straightward <3

PR #75, opened 2025-11-30
Ignore ubsan warnings in libc startup files

dzbarsky:


sanitizers/ubsan_ignore.txt
```diff
@@ -0,0 +1,3 @@
+# Ignore all sanitizer instrumentation for libc startup files.
+src:external/toolchains_llvm_bootstrapped+/third_party/libc/glibc/csu/elf-init-2.31.c
```
cerisier, line 2: this might not work if downstream map repo to a different name ?

PR #87, opened 2025-12-05
Use a custom prefix dir for CRTstuff

cerisier:
This is an attempt to solve #65 as well as a more correct rules_rust support.
But is also a more correct and robust way of dealing with those.

Currently, CRTstuff are passed as `cc_args` manually which causes 2 main issues.

In the case of `rules_go`, link args get duplicated which make the link fail with duplicate symbols.
This is because those need to be always linked. (--whole-archive for libs, or same behavior for direct .o args)

In the case of `rules_rust` (without `experimental_cc_common_link`), the rust linker driver adds rlibs before resolved LD_FLAGS which contains the CRT args.
This has the effect of producing faulty binaries because the order in which those CRT are linked are crucial.

By creating a specific dir, we can now let the clang driver search for the files itself, given they have the exact name it expects (this changes based on `-shared`, `-static`, `-pie`, `-no-pie`).

Essentially, we go from `-nostdlib` (which implies `-nostartfiles`) to `-nodefaultlibs` only.

### DIFF
--- a/runtimes/BUILD.bazel
+++ b/runtimes/BUILD.bazel
@@ -1,8 +1,8 @@
+load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 load("//toolchain:selects.bzl", "platform_llvm_binary")
-load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
-load("@toolchains_llvm_bootstrapped//toolchain/stage2:cc_stage2_object.bzl", "cc_stage2_object")
-load("@toolchains_llvm_bootstrapped//toolchain/args:targets.bzl", "TARGETS")
+load("//toolchain/stage2:cc_stage2_object.bzl", "cc_stage2_object")
+load("//toolchain/args:llvm_target_triple.bzl", "LLVM_TARGET_TRIPLE")
 
 alias(
     name = "llvm_ar",
@@ -51,7 +51,7 @@ cc_stage2_object(
     ],
     copts = [
         "-target",
-    ] + TARGETS,
+    ] + LLVM_TARGET_TRIPLE,
     out = "crti.o",
     visibility = ["//visibility:public"],
 )
@@ -63,7 +63,7 @@ cc_stage2_object(
     ],
     copts = [
         "-target",
-    ] + TARGETS,
+    ] + LLVM_TARGET_TRIPLE,
     out = "crtn.o",
     visibility = ["//visibility:public"],
 )
@@ -92,7 +92,6 @@ copy_to_directory(
         "@compiler-rt//:crtend_object",
         "@compiler-rt//:crtendS_object",
     ],
-    verbose = True,
     include_external_repositories = ["**"],
     target_compatible_with = [
         "@platforms//os:linux",
--- a/toolchain/BUILD.bazel
+++ b/toolchain/BUILD.bazel
@@ -1,4 +1,3 @@
-load("@bazel_skylib//lib:selects.bzl", "selects")
 load("@bazel_skylib//rules:common_settings.bzl", "bool_setting")
 load("@rules_cc//cc/toolchains/impl:documented_api.bzl", "cc_args_list")
 load("@rules_cc//cc/toolchains:args.bzl", "cc_args")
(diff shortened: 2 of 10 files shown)


toolchain/args/BUILD.bazel
```diff
@@ -64,7 +53,7 @@ cc_args(
         "@rules_cc//cc/toolchains/actions:cpp_link_executable",
     ],
     args = [
-        "-static",
+        "-static-pie",
```
cerisier, line 74: I fixed musl in the same time since we were not producing pie executable for musl)

runtimes/BUILD.bazel
```diff
@@ -34,3 +38,66 @@ filegroup(
...
+            "//runtimes/glibc:glibc_Scrt1.object",
+        ],
+    }) + [
+        ":crti_object",
+        ":crtn_object",
+        "@compiler-rt//:crtbegin_object",
+        "@compiler-rt//:crtbeginS_object",
+        "@compiler-rt//:crtbeginT_object",
+        "@compiler-rt//:crtend_object",
+        "@compiler-rt//:crtendS_object",
```
cerisier, line 93: For now I include them all (they are the same file, just named differently), but from my research, if we supported a custom `-resource-dir` we could just put them in a `./lib/` folder as `clang_rt.crtbegin.o` and `clang_rt.crtend.o` and it would resolve just fine.

The added benefit of a custom `-resource-dir` is that it would allow us to put all `clang_rt.*` stuff there (`builtins`, `sanitiers`, etc).

runtimes/BUILD.bazel
```diff
@@ -34,3 +38,68 @@ filegroup(
...
+    ],
+    verbose = True,
```
dzbarsky: snip `verbose=True`?

toolchain/BUILD.bazel
```diff
@@ -53,7 +53,52 @@ cc_args_list(
             "//toolchain/args/macos:macos_default_link_flags",
         ],
         "//conditions:default": [],
-    }, no_match_error = "Unsupported platform for default_link_flags"),
+    }),
+)
+
+cc_args(
```
dzbarsky: should this go under `//tolchain/args`?

toolchain/args/targets.bzl
```diff
@@ -0,0 +1,13 @@
+TARGETS = select({
```
dzbarsky: should this be called LLVM_TARGET_TRIPLE or something?

toolchain/stage2/args/linux/BUILD.bazel
```diff
@@ -0,0 +1,66 @@
+
+load("@rules_cc//cc/toolchains:args.bzl", "cc_args")
+load("@rules_cc//cc/toolchains/impl:documented_api.bzl", "cc_args_list")
+
+package(default_visibility = ["//toolchain/stage2:__subpackages__"])
+
+# Passing a sysroot to /dev/null has the side effect of removing all default
```
dzbarsky: phrasing nit: either `Passing /dev/null as sysroot` or `Passing a sysroot _of_ /dev/null` :)

toolchain/stage2/cc_stage2_object.bzl
```diff
@@ -0,0 +1,75 @@
...
+        action_name = ACTION_NAMES.cpp_link_executable,
+    )
+
+    arguments = ctx.actions.args()
+    arguments.add("-fuse-ld=lld")
+    arguments.add_all(ctx.attr.copts)
+    arguments.add("-r")
+    for src in ctx.files.srcs:
+        if src.path.endswith(".a"):
+            arguments.add_all(["-Wl,--whole-archive", src.path, "-Wl,--no-whole-archive"])
```
dzbarsky: nit: just pass `src`,  no need to `.path` it, same below

toolchain/stage2/cc_stage2_object.bzl
```diff
@@ -0,0 +1,75 @@
...
+    arguments.add("-o")
+    arguments.add(ctx.outputs.out.path)
+
+    ctx.actions.run(
+        inputs = ctx.files.srcs,
+        outputs = [ctx.outputs.out],
+        arguments = [arguments],
+        tools = cc_toolchain.all_files,
+        executable = cc_tool,
+        mnemonic = "CcStage2Compile",
```
dzbarsky, line 49: onced we fix the above and add `execution_requirements = {"supports-path-mapping": "1"}` here I think it should dedupe this action between configs

dzbarsky (approved):

toolchain/stage2/cc_stage2_object.bzl
cerisier: My bad ! Also necessary for path mapping

PR #91, opened 2025-12-06
[windows] Finish up mingw support

dzbarsky:
With this, windows mings amd64/arm64 become valid/working target platforms, and we set up CI to verify things work when cross-compiling back and forth


toolchain/args/llvm_target_triple.bzl
```diff
@@ -8,6 +8,8 @@ LLVM_TARGET_TRIPLE = select({
     "@toolchains_llvm_bootstrapped//platforms/config/libc_aware:linux_aarch64_musl": ["aarch64-linux-musl"],
     "@toolchains_llvm_bootstrapped//platforms/config:macos_x86_64": ["x86_64-apple-darwin"],
     "@toolchains_llvm_bootstrapped//platforms/config:macos_aarch64": ["aarch64-apple-darwin"],
+    "@toolchains_llvm_bootstrapped//platforms/config:windows_x86_64": ["x86_64-w64-windows-gnu"],
```
cerisier, line 11: isn't this x86_64-w64-mingw64 ?

cerisier: Actually no, it's an alias to windows-gnu so all good

third_party/llvm-project/no_frontend_builtin_headers.patch
```diff
@@ -0,0 +1,19 @@
...
+diff --git a/utils/bazel/llvm-project-overlay/clang/BUILD.bazel b/utils/bazel/llvm-project-overlay/clang/BUILD.bazel
+index 5ceb2e42c23d..c7ba4fed1760 100644
+--- a/utils/bazel/llvm-project-overlay/clang/BUILD.bazel
++++ b/utils/bazel/llvm-project-overlay/clang/BUILD.bazel
+@@ -1744,7 +1744,7 @@ cc_library(
+         "include/clang/Frontend/*.h",
+     ]),
+     copts = ["$(STACK_FRAME_UNLIMITED)"],
+-    data = [":builtin_headers_gen"],
++    #data = [":builtin_headers_gen"],
```
cerisier, line 16: IS this new ?

e2e/rules_cc/BUILD.bazel
```diff
@@ -44,3 +57,20 @@ platform_transition_binary(
...
+    # We run CI with `--config=stage1` so we will first compile a Windows-exec-flavored LLVM toolchain from source,
+    # then use that to build this target on the Windows host.
+    exec_compatible_with = [
+        "@platforms//os:windows",
+    ],
+    target_compatible_with = [
+        "@platforms//os:windows",
+    ],
+    linkopts = [
+        "-luuid",
```
cerisier, line 73: is this a test ? or actualy required ? I thought those were only for building clang ?

third_party/llvm-project/no_frontend_builtin_headers.patch
dzbarsky: Its a `genrule` so it doesn't really work well with windows x-build and we don't use it anyway. But also see https://github.com/llvm/llvm-project/blob/dd33690686e69b8fd7802cdd7e875f3e362d268b/utils/bazel/llvm-project-overlay/clang/BUILD.bazel#L1687-L1699 - the `output_to_bindir` is suss, I think this might be hardcoding how google does their toolchain stuff?

e2e/rules_cc/BUILD.bazel
dzbarsky: `-luuid` is needed because the test uses the uuid lib :)
```
    GUID guid = GUID_NULL;
    (void)guid;
```

PR #105, opened 2025-12-10
[windows] Add compiler-rt support

dzbarsky:
See https://github.com/llvm/llvm-project/blob/fec0a64daea642e699c1b31498a17b9718513774/compiler-rt/lib/builtins/CMakeLists.txt#L334

### DIFF
--- a/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
+++ b/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
@@ -237,7 +237,7 @@ filegroup(
         "lib/builtins/fixunsxfti.c",
         "lib/builtins/floatdixf.c", # if not android
         "lib/builtins/floattixf.c",
-        "lib/builtins/floatundixf.c", # if not win32
+        "lib/builtins/floatundixf.c",
         "lib/builtins/floatuntixf.c",
         "lib/builtins/mulxc3.c",
         "lib/builtins/powixf2.c",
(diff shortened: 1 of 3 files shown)


third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
```diff
@@ -237,7 +237,7 @@ filegroup(
         "lib/builtins/fixunsxfti.c",
         "lib/builtins/floatdixf.c", # if not android
         "lib/builtins/floattixf.c",
-        "lib/builtins/floatundixf.c", # if not win32
+        "lib/builtins/floatundixf.c",
```
cerisier, line 240: this will get built for windows then ?

third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
```diff
@@ -422,13 +441,17 @@ cc_stage2_library(
         "@platforms//os:linux": [
             "@kernel_headers//:kernel_headers",
         ],
+        "@platforms//os:windows": [
+            # "@mingw//:mingw_headers",
```
cerisier, line 445: \# ?

dzbarsky: I haven't added that target yet but this is where it would go :) I can remove for now if you prefer

third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
dzbarsky: yeah i wasn't sure why you added this comment, see https://github.com/llvm/llvm-project/blob/fec0a64daea642e699c1b31498a17b9718513774/compiler-rt/lib/builtins/CMakeLists.txt#L315

PR #108, opened 2025-12-11
[windows] Silence mingw warnings in runtime libs

dzbarsky:


runtimes/mingw/crt_object.bzl
```diff
@@ -0,0 +1,22 @@
+load("@bazel_lib//lib:copy_file.bzl", "copy_file")
+load("@toolchains_llvm_bootstrapped//toolchain/stage2:cc_stage2_library.bzl", "cc_stage2_library")
+
+def crt_object(name, out, visibility, **kwargs):
```
cerisier, line 4: unused ?

dzbarsky: It will be used in future PR and snuck its way in, might as well keep :)

PR #110, opened 2025-12-11
[windows] Build runtime CRT libs

dzbarsky:


runtimes/mingw/crt_sources.bzl
```diff
@@ -0,0 +1,545 @@
...
+    "math/sincos.c",
+    "math/sincosf.c",
+    "math/arm-common/sincosl.c",
+    "math/arm64/rint.c",
+    "math/arm64/rintf.c",
+    "math/arm-common/ldexpl.c",
+]
+
+LIBM_SRCS = [
+    "_libm_dummy.c",
```
cerisier: this is a generated file ?

dzbarsky: Yeah this bit shouldn't exist. I thought I snipped it. It's generated file

PR #113, opened 2025-12-13
[windows] Generate import libs

dzbarsky:


runtimes/mingw/BUILD.tpl
```diff
@@ -132,3 +141,25 @@ write_file(
...
+    directory = "mingw-w64-crt/lib64",
+)
+
+directory(
+    name = "mingw_import_libraries_directory",
+    srcs = [
+        ":mingw_import_libraries_common",
+    ] + select({
+        "@platforms//cpu:x86_64": [":mingw_import_libraries_x86_64"],
+        # Aarch64 are fully templatized in .def.in files, x86_64 have not been merged yet.
```
cerisier, line 161: Not sure I fully understand what it implies !

dzbarsky: The common folder has .def.in files that run thru preprocessor to generate the arch-specific .def

So arm64 is all handled generically this way. There's no arm64 specific .def files. But there are for x86

PR #114, opened 2025-12-13
[windows] Wire up default toolchain flags

dzbarsky:


toolchain/args/windows/BUILD.bazel
```diff
@@ -0,0 +1,66 @@
...
+    ],
+    args = [
+        "-L{mingw_import_library_search_path}",
+        "-L{mingw_crt_library_search_path}",
+
+        "-nodefaultlibs",
+
+        "-lmingw32",
+        "-lmoldname",
+        "-lmsvcrt",
```
cerisier: I think this one is not correct as I think we want to explicitly avoid -lmsvcrt ? or at least not have both -lmsvcrt and -lucrt ?

From a clang invocation with a mingw64 sysroot i see:

-lucrt
-lmingw32
-lgcc
-lgcc_eh
-lmoldname
-lmingwex
-ladvapi32
-lshell32
-luser32
-lkernel32

in this exact order.

How did you find those at of interest ?

dzbarsky: I tried something along those lines but saw linker errors without `lole32` on LLVM binaries. Let me double check though!

dzbarsky: oh and yes there should not be `lmsvcrt` you are correct :)

dzbarsky: So I get the following linker errors:
```
ld.lld: error: undefined symbol: __declspec(dllimport) CoInitializeEx
>>> referenced by libSupport.a(Path.o):(llvm::sys::fs::remove_directories(llvm::Twine const&, bool))

ld.lld: error: undefined symbol: __declspec(dllimport) CoCreateInstance
>>> referenced by libSupport.a(Path.o):(llvm::sys::fs::remove_directories(llvm::Twine const&, bool))

ld.lld: error: undefined symbol: __declspec(dllimport) CoUninitialize
>>> referenced by libSupport.a(Path.o):(llvm::sys::fs::remove_directories(llvm::Twine const&, bool))

ld.lld: error: undefined symbol: __declspec(dllimport) CoTaskMemFree
>>> referenced by libSupport.a(Path.o):(llvm::sys::path::getKnownFolderPath(_GUID, llvm::SmallVectorImpl<char>&))

ld.lld: error: undefined symbol: CLSID_FileOperation
>>> referenced by libSupport.a(Path.o):(.refptr.CLSID_FileOperation)

ld.lld: error: undefined symbol: FOLDERID_LocalAppData
>>> referenced by libSupport.a(Path.o):(.refptr.FOLDERID_LocalAppData)

ld.lld: error: undefined symbol: FOLDERID_Profile
>>> referenced by libSupport.a(Path.o):(.refptr.FOLDERID_Profile)
```

These necessitate adding `-luuid` and `-lole32`. Actually though the bug is in LLVM's BUILD files, see https://github.com/llvm/llvm-project/blob/fa1dceb67f6fef652c2bf91c070ede64eadc49c6/llvm/lib/Support/CMakeLists.txt#L44.

So I can remove them here and add it to my "make LLVM on windows behave" upstream PR :)

dzbarsky: ok i rolled the extra libs into https://github.com/cerisier/toolchains_llvm_bootstrapped/pull/115

PR #115, opened 2025-12-13
[windows] Add LLVM patches to make windows cross-builds work correctly

dzbarsky:
This fixes a few issues:
- genrule is broken because it bakes host bash path into the command which gets sent to remote executor. This can't possibly work cross-platform
- Upstream assumed clang-cl style always, which we aren't using. Harmonize the flags with what is actually expected
- Few missing libs link flags in llvm/lib/Support
- py_binary is also annoying because it has a bash dep on the bootstrap. We can just rewrite to C, it's pretty trivial.

The first one is already being upstreamed at https://github.com/llvm/llvm-project/pull/171761
The second one might be more contentious (might not want to diverge cmake and bazel) so I'm holding off until we get the first one merged - this one is also much less annoying for us to carry since it's so self-contained

### DIFF
--- a/third_party/llvm-project/windows_link_and_genrule.patch
+++ b/third_party/llvm-project/windows_link_and_genrule.patch
@@ -1,0 +1,373 @@
+commit e0b69f7d1737a96728b5918e121f6a4057c45437
+Author: David Zbarsky <dzbarsky@gmail.com>
+Date:   Mon Dec 8 13:55:15 2025 -0500
+
+    No more genrules
+
+diff --git a/llvm/lib/Support/BLAKE3/blake3_avx2_x86-64_unix.S b/llvm/lib/Support/BLAKE3/blake3_avx2_x86-64_unix.S
+index e98893c7ef8b..d6fa51638da3 100644
+--- a/llvm/lib/Support/BLAKE3/blake3_avx2_x86-64_unix.S
++++ b/llvm/lib/Support/BLAKE3/blake3_avx2_x86-64_unix.S
+@@ -16,15 +16,17 @@
+ #define _CET_ENDBR
+ #endif
+ 
+-#ifdef __APPLE__
+-#define HIDDEN .private_extern
++#if defined(__APPLE__)
++#define HIDDEN(symbol) .private_extern symbol
++#elif defined(__ELF__)
++#define HIDDEN(symbol) .hidden symbol
+ #else
+-#define HIDDEN .hidden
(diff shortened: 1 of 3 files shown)


third_party/llvm-project/windows_link_and_genrule.patch
```diff
@@ -0,0 +1,373 @@
...
++    content = [
++        "// extension handlers",
++        "#undef HANDLE_EXTENSION",
++    ],
++)
++
++config_setting(
++    name = "is_windows_clang_mingw",
++    constraint_values = ["@platforms//os:windows"],
++    flag_values = {"@rules_cc//cc/compiler:compiler": "clang"},
```
cerisier, line 335: Doesn't really imply mingw does it ?

third_party/llvm-project/windows_link_and_genrule.patch
```diff
@@ -0,0 +1,373 @@
...
+@@ -297,7 +320,21 @@ cc_library(
+     }),
+     includes = ["include"],
+     linkopts = select({
+-        "@platforms//os:windows": [
++        ":is_windows_clang_mingw": [
++            "-lole32",
++            "-luuid",
++            "-lws2_32",
++            "-lntdll",
```
cerisier, line 360: I will investigate why we need this

third_party/llvm-project/windows_link_and_genrule.patch
dzbarsky: this is specifically clang and not clang-cl. I think mingw is the only real option aside from clang in msvc mode and msvc itself :)

I suppose it could be renamed to not be "mingw" technically, but meh. Lets go with whatever upstream accepts

third_party/llvm-project/windows_link_and_genrule.patch
dzbarsky: See https://github.com/llvm/llvm-project/blob/fa1dceb67f6fef652c2bf91c070ede64eadc49c6/llvm/lib/Support/CMakeLists.txt#L44C48-L44C53

I don't know how this ever built without it...

third_party/llvm-project/windows_link_and_genrule.patch
cerisier: There is cygwin theoritically haha by yeah not important

dzbarsky: oh, right. 🙃

PR #122, opened 2025-12-14
Simplify mtree construction

dzbarsky:


prebuilt/llvm/llvm_release.bzl
```diff
@@ -34,60 +20,46 @@ def llvm_release(name):
         tags = ["manual"],
     )
 
-    native.genrule(
-        name = "sanitizers_ignorelist_mtree",
```
cerisier, line 38: Not anymore ?

cerisier (approved):

prebuilt/llvm/llvm_release.bzl
dzbarsky: I just moved this into `bins_mtree` (even though it's slightly misnamed) because it doesn't really clutter things nearly as much anymore :)

PR #124, opened 2025-12-14
Enable ThinLTO feature for linux

dzbarsky:


dzbarsky:
> Are those patches upstreamable to rules_cc ?

they already landed upstream weeks 1-2 weeks ago, we just need a release...

PR #125, opened 2025-12-15
Use a link only resource-dir

cerisier:
This is primarily so that go tool link linkerFlagSupported checks, which trims a lot of flags, can pass.
But actually this is also the best strategy for our toolchain for now, because we should not mix half driver, half not driver.
Additionally, this will also make it easier to generate cross toolchains for usage outside of bazel.

### DIFF
--- a/toolchain/args/BUILD.bazel
+++ b/toolchain/args/BUILD.bazel
@@ -234,7 +234,6 @@ cc_args(
         "@rules_cc//cc/toolchains/actions:link_actions",
     ],
     args = [
-        # "-L.", # needed to -l:bazel-out/...
         "-L{libcxx_library_search_path}",
         "-L{libunwind_library_search_path}",
     ],


runtimes/compiler-rt/BUILD.bazel
```diff
@@ -4,6 +4,12 @@ alias(
     visibility = ["//visibility:public"],
 )
 
+alias(
```
dzbarsky, line 7: do you think these aliases are actually worth having? Given `copy_to_resource_directory` configures the final file names anyway, feels like having more slightly different names for the same thing just adds confusion/indirection?

runtimes/copy_to_resource_directory.bzl
```diff
@@ -0,0 +1,81 @@
...
+    doc = "Copies the given srcs into a resource directory layout under lib/<triple>/.",
+    implementation = _copy_to_resource_directory_rule_impl,
+    attrs = {
+        "srcs": attr.label_keyed_string_dict(
+            doc = "Dict of label -> basename. Each value is the filename to appear under lib/<triple>/",
+            mandatory = True,
+            allow_files = True,
+        ),
+        "target_triple": attr.string(
+            doc = "The target triple to use for placing the files.",
```
dzbarsky, line 62: this is also technically mandatory right?

toolchain/args/BUILD.bazel
```diff
@@ -217,19 +234,16 @@ cc_args(
         "@rules_cc//cc/toolchains/actions:link_actions",
     ],
     args = [
-        "-L.", # needed to -l:bazel-out/...
+        # "-L.", # needed to -l:bazel-out/...
```
dzbarsky: snip?

toolchain/args/linux/BUILD.bazel
```diff
@@ -37,12 +37,17 @@ cc_args(
 )
 
 cc_args(
+    # We provide the libc++ and libunwind ourselves through static_runtime_libs
+    # and dynamic_runtime_libs.
     name = "hermetic_link_flags",
```
dzbarsky, line 42: maybe this should be renamed to runtime_lib_flags or something?

toolchain/args/linux/BUILD.bazel
```diff
@@ -120,14 +125,16 @@ cc_args(
...
+        # "-lc",
+        # Since libunwind is linked statically, we need to link dl as well.
         "-ldl",
-        "-lrt",
-        "-lutil",
+        # "-lrt",
+        # "-lutil",
 
-        "-Wl,--pop-state",
+        # "-Wl,--pop-state",
```
dzbarsky: don't we still need the pop at the end? I feel like this may be working by accident if these flags are getting set last?

toolchain/args/linux/BUILD.bazel
```diff
@@ -120,14 +125,16 @@ cc_args(
         # But in the way we handle generating libc stubs, we chose to generate
         # empty libs for those that have been removed so that we do not have to
         # handle varying inclusions.
-        "-lm",
+        # "-lm",
+        # Since libunwind is linked statically, we need to link pthread as well.
         "-lpthread",
-        "-lc",
+        # "-lc",
```
dzbarsky: remove the libc directives?

runtimes/compiler-rt/BUILD.bazel
cerisier: I liked having a whole range of targets available at //runtimes so that i can predictably refer to them.
Also, the initial idea was to be able to select the llvm version haha

dzbarsky: Oh right I forgot about that. Ok let's keep it, maybe add a note so we don't forget

runtimes/copy_to_resource_directory.bzl
cerisier: I didn't find another way to select on that :( default doesn't accept a select.

toolchain/args/BUILD.bazel
cerisier: oh yes!

toolchain/args/linux/BUILD.bazel
cerisier: good catch it was a mistake.

toolchain/args/linux/BUILD.bazel
cerisier: yep

PR #126, opened 2025-12-15
Enable windows cross-compilation test suite

dzbarsky:


MODULE.bazel
```diff
@@ -191,3 +191,5 @@ register_toolchains(
     "@zig_toolchains//:all",
     dev_dependency = True,
 )
+
+bazel_dep(name = "ape", version = "1.1.0")
```
cerisier: Do you think this dep is acceptable by everyone ?

dzbarsky: Honestly this module seems very overengineered, I don't understand why downloading a prebuilt binary is so complex. We can probably strip all of it away, just wanted to get it working first.

BTW, there is `@sed` in BCR but we will have a bootstrap problem using it for musl...should we distribute it as extra prebuilts?

cerisier: We can yes !

PR #131, opened 2025-12-15
Rework the way we declare platforms configs and constraints

cerisier:
Got rid of the libc_aware package and merged it with the normal declarations.

Removed the unconstrained libc variant, the default is now gnu.2.28 and is set both as default value so that platforms from users not referencing the constraint still get the default value, and also set as part of our own platform definitions so we have  a explicit vision of which glibc is the default.

Refined much config in selects to use the narrowest constraint (@platforms//cpu:aarch64 instead of linux_aarch64), avoiding repetitions for all targets we will support.

### DIFF
--- a/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
+++ b/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
@@ -314,7 +314,7 @@ cc_stage2_library(
         "@platforms//cpu:x86_64": [":builtins_x86_64_sources"],
         "@platforms//cpu:aarch64": [":builtins_aarch64_sources"],
     }, no_match_error = """
-        Platform not supported for compiler-rt.builtins.
+        Architecture not supported for compiler-rt.builtins.
         It is likely that we are just missing the filegroups for that platform.
         Please file an issue.
     """) + [


platforms/config/declare_config_settings.bzl
```diff
@@ -10,3 +12,44 @@ def declare_config_settings():
...
+                name = "{}_{}_{}".format(target_os, target_cpu, libc),
+                constraint_values = [
+                    "@platforms//cpu:{}".format(target_cpu),
+                    "@platforms//os:{}".format(target_os),
+                    "//constraints/libc:{}".format(libc),
+                ],
+                visibility = ["//visibility:public"],
+            )
+
+        selects.config_setting_group(
```
dzbarsky, line 31: i'm a little bit worried about what this expands into (shitload of aliases) but we can optimize it later

runtimes/BUILD.bazel
```diff
@@ -72,15 +72,11 @@ cc_stage2_object(
...
+        "//platforms/config:musl": [
             # -static-pie is implied
             "//runtimes/musl:musl_rcrt1.object",
         ],
-        "//platforms/config/libc_aware:gnu": [
-            "//runtimes/glibc:glibc_Scrt1.object",
-            "//runtimes/glibc:glibc_crt1.object",
-        ],
-        "//constraints/libc:unconstrained": [
+        "//platforms/config:gnu": [
```
dzbarsky, line 79: yeah this is way more sane :)

third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
```diff
@@ -311,12 +311,8 @@ cc_stage2_library(
...
     includes = ["lib/builtins"],
     srcs = select({
-        "@toolchains_llvm_bootstrapped//platforms/config:linux_x86_64": [":builtins_x86_64_sources"],
-        "@toolchains_llvm_bootstrapped//platforms/config:linux_aarch64": [":builtins_aarch64_sources"],
-        "@toolchains_llvm_bootstrapped//platforms/config:macos_x86_64": [":builtins_x86_64_sources"],
-        "@toolchains_llvm_bootstrapped//platforms/config:macos_aarch64": [":builtins_aarch64_sources"],
-        "@toolchains_llvm_bootstrapped//platforms/config:windows_x86_64": [":builtins_x86_64_sources"],
-        "@toolchains_llvm_bootstrapped//platforms/config:windows_aarch64": [":builtins_aarch64_sources"],
+        "@platforms//cpu:x86_64": [":builtins_x86_64_sources"],
+        "@platforms//cpu:aarch64": [":builtins_aarch64_sources"],
```
dzbarsky, line 315: tweak the mesage below this? platform -> arch

PR #135, opened 2025-12-15
Support compiler-rt for wasm

cerisier:
This is a niche usecase but with an explicit -resource-dir it means they are picked up by default :)

### DIFF
--- a/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
+++ b/third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
@@ -462,22 +462,22 @@ cc_stage2_library(
         "@platforms//os:linux": [
             "@kernel_headers//:kernel_headers",
         ],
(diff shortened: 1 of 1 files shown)


third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
```diff
@@ -441,20 +462,22 @@ cc_stage2_library(
         "@platforms//os:linux": [
             "@kernel_headers//:kernel_headers",
         ],
-        "@platforms//os:windows": [
```
dzbarsky, line 444: How did you know to flip the order?

cerisier: I think it's similar but I guess I consider mingw-headers closer to libc headers than kernel headers.
btw, I need to rename kernel-headers to linux-headers in this case.

PR #136, opened 2025-12-16
[compiler-rt] Simplify darwin sources filtering

dzbarsky:


third_party/llvm-project/20.x/compiler-rt/BUILD.tpl
```diff
@@ -190,7 +189,7 @@ copy_file(
 # buildifier: disable=constant-glob
 filegroup(
```
cerisier, line 190: shouldn't thoe become filter_builtin_sources ?

dzbarsky: this filegroup is itself consumed in `filter_builtin_sources` targets. Like so:

```
filter_builtin_sources(
    name = "builtins_x86_64_sources",
    srcs = [
        ":builtins_generic_srcs",
        ":builtins_generic_tf_sources",
        ":builtins_x86_arch_sources",
```

so the rule of thumb is to do all filtering in single place at top-level before feeding the sources into cc_stage2_library!

PR #149, opened 2025-12-19
Do not pass musl headers as -isystem when building musl itself

cerisier:
This results in weird linkage error most probably because musl includes some of its own headers as "..."

### DIFF
--- a/kernel/extension/BUILD.kernel-headers.tpl
+++ b/kernel/extension/BUILD.kernel-headers.tpl
@@ -10,6 +10,7 @@ cc_library(
     includes = [
         "include",
     ],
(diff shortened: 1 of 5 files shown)


third_party/libc/glibc/BUILD.tpl
```diff
@@ -287,12 +287,8 @@ cc_stage2_library(
         "include/libc-modules.h",
         "include/libc-symbols.h",
     ],
-    implementation_deps = select({
-        "@platforms//os:macos": [],
-        "@platforms//os:linux": [
-            ":kernel_headers",
-        ],
-    }) + [
+    implementation_deps = [
```
dzbarsky, line 290: lol

PR #162, opened 2025-12-27
Stop using a dedicated cc_toolchain for stage2

cerisier:
We now use the transition to choose the correct toolchain args within a single toolchain.
I also took the opportunity to clean a bit the args names and usages.

I want to do a bigger revamp of whole naming and fix actualy default args but thought I'd do that in multiple passes !

### DIFF
--- a/toolchain/args/linux/BUILD.bazel
+++ b/toolchain/args/linux/BUILD.bazel
@@ -51,7 +51,7 @@ cc_args(
     ],
 )
 
-# Extracted from rules_cc unix_cc_configure.bz;
+# Extracted from rules_cc unix_cc_configure.bzl
 cc_args(
     name = "default_link_flags",
     actions = [
--- a/toolchain/cc_toolchain.bzl
+++ b/toolchain/cc_toolchain.bzl
@@ -72,11 +72,11 @@ def cc_toolchain(name, tool_map):
         }),
         tool_map = tool_map,
         static_runtime_lib = select({
-            "//toolchain:bootstrapping": "//runtimes:empty_filegroup",
+            "//toolchain:bootstrapping": "//runtimes:none",
             "//conditions:default": "//runtimes:static_runtime_lib",
         }),
         dynamic_runtime_lib = select({
-            "//toolchain:bootstrapping": "//runtimes:empty_filegroup",
+            "//toolchain:bootstrapping": "//runtimes:none",
             "//conditions:default": "//runtimes:dynamic_runtime_lib",
         }),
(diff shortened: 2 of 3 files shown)


toolchain/args/linux/BUILD.bazel
```diff
@@ -51,6 +51,18 @@ cc_args(
     ],
 )
 
+# Extracted from rules_cc unix_cc_configure.bz;
```
dzbarsky: .bzl filename typo

toolchain/cc_toolchain.bzl
```diff
@@ -21,22 +43,41 @@ def cc_toolchain(name, tool_map):
...
+        }),
         tool_map = tool_map,
-        static_runtime_lib = "//runtimes:static_runtime_lib",
-        dynamic_runtime_lib = "//runtimes:dynamic_runtime_lib",
+        static_runtime_lib = select({
+            "//toolchain:bootstrapping": "//runtimes:empty_filegroup",
+            "//conditions:default": "//runtimes:static_runtime_lib",
+        }),
+        dynamic_runtime_lib = select({
+            "//toolchain:bootstrapping": "//runtimes:empty_filegroup",
```
dzbarsky: Maybe it would read better as //runtimes:none ?

toolchain/stage2/args/linux/BUILD.bazel
```diff
@@ -4,20 +4,7 @@ load("@rules_cc//cc/toolchains/impl:documented_api.bzl", "cc_args_list")
 
 package(default_visibility = ["//toolchain/stage2:__subpackages__"])
 
-# Passing `/dev/null` as sysroot has the side effect of removing all default
```
dzbarsky, line 7: We're sure this doesn't result in any new non-hermetic search paths?

toolchain/cc_toolchain.bzl
cerisier: correct. i'll fix.

toolchain/stage2/args/linux/BUILD.bazel
cerisier: I didn't remove it actually. I just reused the one from //toolchain.
With our setup, the only thing that --sysroot=/dev/null does (but still important), is to remove -L/usr/lib that clang auto adds to linker args.

dzbarsky: Ah ok, sorry I missed it. And yeah I was worried about those auto-add args, thanks for confirming!

PR #163, opened 2025-12-27
Factorize toolchains definition

cerisier:
Organise toolchain args as a group of semantic args list.
Do this for runtimes as well.
Factorize and reuse as much as possible between the 2.

### DIFF
--- a/toolchain/args/BUILD.bazel
+++ b/toolchain/args/BUILD.bazel
@@ -235,10 +235,10 @@ cc_args(
         "@rules_cc//cc/toolchains/actions:link_actions",
     ],
     args = select({
-        ":empty_sysroot": [],
-        "//conditions:default": [
+        ":empty_sysroot": [
             "--sysroot=/dev/null",
         ],
+        "//conditions:default": [],
     }),
 )
 


toolchain/BUILD.bazel
```diff
@@ -101,18 +102,54 @@ cc_args_list(
...
+    name = "sysroot_flags",
+    args = select({
+        "@platforms//os:macos": [
+            "//toolchain/args/macos:macos_sdk_sysroot",
+        ],
+        "//conditions:default": [
+            "//toolchain/args:empty_sysroot_flags",
+        ],
+    }),
+    visibility = ["//toolchain:__subpackages__"],
```
dzbarsky, line 115: should we makethis default package vis?

toolchain/args/BUILD.bazel
```diff
@@ -218,6 +218,45 @@ cc_args(
...
+# Passing a sysroot to /dev/null has the side effect of removing all default
+# library search paths that clang passes like -L/usr/lib.
+cc_args(
+    name = "empty_sysroot_flags",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+        "@rules_cc//cc/toolchains/actions:link_actions",
+    ],
+    args = select({
+        ":empty_sysroot": [],
```
dzbarsky: isn't this backward? lets just kill the config_setting?

cerisier: I didn't want to remove it just yet because I think there is 1 case with rust std where i had no choice but do this ...

toolchain/BUILD.bazel
cerisier: We can't because of config_settings just yet. but yes we should aim at that.

toolchain/args/BUILD.bazel
dzbarsky: Ah ok but it seems weird that passing the empty flag doesn't pass /dev/null, is it backwards or confusingly named?

cerisier: aaa sorry !

cerisier: Very good catch thanks.

PR #165, opened 2025-12-27
Use source directories for builtin_headers

dzbarsky:


dzbarsky:
> This is a workaround while we wait for keith commit ?

yes, but also maybe we just keep it like this, it seems fine

PR #171, opened 2025-12-30
Restructure toolchain target settings

cerisier:
The idea is to simplify the mental model of toolchain resolution, especially as we bring "stage1" (to be renamed).

In this PR:
1. We change config settings names to better match the intents
2. We change the matching strategy to be explicit vs fallback

For 1:

Which runtimes do we need? (no runtimes? + CRTs? +libc? +libc++?)
Which compiler should be used? (prebuilt? from source aka bootstrapped?)

For 2:

We start with default build settings value for both:
- all runtimes
- prebuilt

And we constraint toolchain declarations to either prebuilt or bootstrapped.

This allows regstering all toolchains, and using stage1 by using a build flag.
```
common:stage1   --@toolchains_llvm_bootstrapped//toolchain:compiler_flavor=bootstrapped
```

### DIFF
--- a/toolchain/runtimes/cc_stage0_object.bzl
+++ b/toolchain/runtimes/cc_stage0_object.bzl
@@ -51,7 +51,7 @@ def _cc_stage0_object_impl(ctx):
         tools = cc_toolchain.all_files,
         executable = cc_tool,
         execution_requirements = {"supports-path-mapping": "1"},
-        mnemonic = "Ccstage0Compile",
+        mnemonic = "CcStage0Compile",
     )
 
     return [DefaultInfo(files = depset([ctx.outputs.out]))]


toolchain/BUILD.bazel
```diff
@@ -1,47 +1,75 @@
...
+        # "stage1", # depends on above: libunwind
+        # "stage2", # depends on above: libc++abi
+        # "stage3", # depends on above: libc++
+        "complete", # depends on above: user programs.
+    ],
+    build_setting_default = "complete",
+)
+
+string_flag(
+    name = "compiler_flavor",
```
cerisier: Could also be named:
- `toolchain_flavor`
- `toolchain_provisioning_mode`
- `toolchain_mode`
- `toolchain_artifact_origin`

toolchain/BUILD.bazel
```diff
@@ -1,47 +1,75 @@
...
 load(":declare_toolchains.bzl", "declare_toolchains")
 
-bool_setting(
-    name = "bootstrap_setting",
-    build_setting_default = False,
+# TODO(cerisier): Find a better name ?
+# This should be the semantic for "what are we building ?"
+# Runtimes without dependencies, runtimes with dependencies, user programs with runtime dependencies...
+string_setting(
+    name = "runtime_stage",
```
cerisier, line 13: Could also be named:
- `runtime_dependencies`
- `runtime_inclusion`

Any other idea ?

.bazelrc
```diff
@@ -14,6 +14,7 @@ common --repo_env=BAZEL_NO_APPLE_CPP_TOOLCHAIN=1
 common --experimental_cc_static_library
 common --experimental_platform_in_output_dir
 
+common:stage1   --@toolchains_llvm_bootstrapped//toolchain:compiler_flavor=bootstrapped
 common:stage1   --extra_toolchains=@toolchains_llvm_bootstrapped//toolchain/stage1:all
```
cerisier: We can remove this if we register `/...`.

toolchain/BUILD.bazel
```diff
@@ -1,47 +1,75 @@
...
+    values = [
+        "stage0", # depend on compiler-only: CRTs, builtins, libc, static libraries.
+        # "stage1", # depends on above: libunwind
+        # "stage2", # depends on above: libc++abi
+        # "stage3", # depends on above: libc++
+        "complete", # depends on above: user programs.
```
cerisier, line 19: I added those as stages because later, we will need to break dependency cycles for instance when building `libc++.so` which needs to depend on `libc.so, crt and compiler-rt`.

toolchain/BUILD.bazel
dzbarsky: maybe `source` = bootstrapped / prebuilt? too terse?

toolchain/BUILD.bazel
dzbarsky: maybe just `stage`? I think that's fairly standard terminology for toolchain bootstrap and this is really internal, right? Or do we expect users to use this as well potentially? (Should we make a `//private` at some point so we can cleanup public/private API?)

cerisier: It's not really a toolchain stage naming. It's just the runtimes requirements.

toolchain/runtimes/cc_unsanitized_library.bzl
```diff
@@ -5,10 +5,17 @@ def _reset_sanitizers_impl(settings, attr):
         "//config:ubsan": False,
         "//config/stage1:ubsan": False,
 
+        # we are compiling sanitizers, so we want all runtimes except sanitizers.
+        # TODO(cerisier): Should this be exressed with a dedicated stage ?
+        "//toolchain:runtime_stage": "complete",
```
dzbarsky, line 10: likely yes, since they depend on llvm libs which depend on the other runtime components

toolchain/BUILD.bazel
dzbarsky: maybe `runtime_components`?

README.md
```diff
@@ -29,7 +29,7 @@ Add this to your `MODULE.bazel`:
 bazel_dep(name = "toolchains_llvm_bootstrapped", version = "0.3.1")
 
 register_toolchains(
-    "@toolchains_llvm_bootstrapped//toolchain:all",
+    "@toolchains_llvm_bootstrapped//toolchain/...",
```
dzbarsky: i'm marginally worried about this because if we get cosmo/ape-exec toolchains and expose them in `//toolchain/cosmo` , this will register them as well as the normal ones, which is probably not what we want. Perhaps we can use some aliases to expose the bootstrapped toolchains in the `//toolchain:all` package or something? Not sure of a good way to do it, just don't want to churn users into using `/...` and then having to potentially undo it

cerisier: Makes sense !

toolchain/runtimes/README.md
```diff
@@ -0,0 +1,6 @@
+# Runtime only toolchain
+
+This package holds the toolchain args for compiling runtimes.
+
+This it not a dedicated toolchain, but targets of the cc_args family used by
```
dzbarsky: "This it" -> "This is"

toolchain/runtimes/cc_stage0_common.bzl
```diff
@@ -1,21 +1,23 @@
...
+    # We are compiling runtimes without any kind of other dependencies.
     builder.set(
-        Label("//toolchain:bootstrap_setting"),
-        True,
+        Label("//toolchain:runtime_stage"),
+        "stage0",
     )
 
+    # TODO(cerisier): Why constraint here ?
     builder.set(
```
dzbarsky, line 20: This basically prevents an extra build of the runtime libs, but is not technically correct :)

toolchain/runtimes/cc_stage0_object.bzl
```diff
@@ -50,14 +51,14 @@ def _cc_stage2_object_impl(ctx):
         tools = cc_toolchain.all_files,
         executable = cc_tool,
         execution_requirements = {"supports-path-mapping": "1"},
-        mnemonic = "CcStage2Compile",
+        mnemonic = "Ccstage0Compile",
```
dzbarsky: uber-nit: Ccstage0Compile -> CcStage0Compile :)

PR #174, opened 2026-01-06
Support dynamic linking

### LINKED ISSUE #23: Support dynamic linking for the C++ standard library
As soon as we link against shared libraries in a c++ project, the c++ standard library should be linked dynamically.
This issue tracks all sub issues.

cerisier:
Fixes #23 altough only for linux for now.

This PR brings an important change.

By default, `linkstatic=True` for `cc_binary` targets.
By honoring the `linkstatic` attribute through `static_runtime_lib` and `dynamic_runtime_lib`, this means that by default, `cc_binary` targets will link against static cpp runtimes.

This means that if one wants dynamic cpp runtimes, they'll have to add `linkstatic=False` to their targets.

This basically makes this toolchain compliant with the specs:

https://github.com/bazelbuild/rules_cc/blob/640996472ed33723ce71347f99509b52f84118b2/cc/private/rules_impl/attrs.bzl#L236-L247

Additionally we had to:
- Introduce a stage1 runtime shared library so we can link libc++.so etc against the libc.so
- force add -rpath $ORIGIN for runtime library since rules_cc refers to google specific $EXEC_ORIGIN for cc_test

### DIFF
--- a/e2e/rules_cc/BUILD.bazel
+++ b/e2e/rules_cc/BUILD.bazel
@@ -57,7 +57,6 @@ cc_test(
         "//conditions:default": [],
     }),
     exec_compatible_with = HOST_CONSTRAINTS,
-    target_compatible_with = HOST_CONSTRAINTS,
     # remote-exec tests can't run without forcing the target platform to match
     # the remote exec platform.
     tags = ["no-remote"],


runtimes/glibc/BUILD.bazel
```diff
@@ -82,6 +83,26 @@ write_file(
     ],
 )
 
+# Same as above.
+write_file(
+    name = "__tls_get_addr.S",
```
cerisier, line 88: The real story involves linking against the dynamic linker using the linker script. BUT this implies generating the correct ld.so name and duplicates part of the clang driver job which I'd like to avoid as long as possible ?

runtimes/musl/BUILD.bazel
```diff
@@ -71,13 +72,44 @@ cc_stage0_library(
...
+        ":musl_ldso",
+    ],
+    user_link_flags = [
+        "-Wl,-e,_dlstart",
+        # should it be "now"?
+        "-Wl,-z,now",
+        "-Wl,-soname,libc.so",
+    ],
+    visibility = ["//visibility:public"],
+)
```
cerisier, line 110: Unused for now.

runtimes/BUILD.bazel
```diff
@@ -3,6 +3,34 @@ load("@rules_cc//cc:cc_library.bzl", "cc_library")
...
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "linkmode_dynamic",
+    flag_values = {
+        ":linkmode": "dynamic",
+    },
+    visibility = ["//visibility:public"],
+)
```
cerisier, line 33: Used to choose compile flags for runtimes between static and dynamic (mostly symbol visibility)

runtimes/BUILD.bazel
```diff
@@ -35,9 +62,7 @@ filegroup(
         "@platforms//os:linux": [
             "@libcxx//:libcxx.shared",
             "@libcxxabi//:libcxxabi.shared",
-            # only if libcxx on linux
-            # TODO(zbarsky): Enable this? libunwind depends on a another shared lib which loses GraphNodeInfo provider
-            #"@libunwind//:libunwind.shared",
+            "@libunwind//:libunwind.shared",
```
cerisier, line 65: This is solved by depending on a `cc_library` and not a `cc_stage0_library`.
I think I will do a pass and replace all `cc_stage0_library` with `cc_library` inside `//third_party` and only refer to those in the main repo.

runtimes/glibc/BUILD.bazel
```diff
@@ -54,16 +54,17 @@ select_file(
...
-# Because we let clang frontend provide the -dynamic-linker flag, it means that
-# the linker will never be able to link against the actual target dynamic linker,
-# only set the proper path as the dynamic linker.
+# To avoid having to compute the proper ld.so name in the linker script, we
+# resort to providing those symbols here.
 #
-# Because of this, references to __stack_chk_guard would end up as undefined
-# symbol. Here we add a stub in the libc because we know that the
-# dynamic linker will be the one to provide it.
+# These symbols may not be normally availble for all targets and versions of the
```
dzbarsky: nit: availble typo

runtimes/musl/BUILD.bazel
```diff
@@ -99,7 +131,7 @@ cc_stage0_library(
     visibility = ["//visibility:public"],
 )
 
-cc_stage0_library(
+cc_library(
```
dzbarsky, line 134: how do we avoid leaking --copt now?

third_party/llvm-project/20.x/libcxx/BUILD.tpl
```diff
@@ -1224,12 +1224,16 @@ cc_library(
         "_LIBCPP_REMOVE_TRANSITIVE_INCLUDES",
         "LIBCXX_BUILDING_LIBCXXABI",
         "_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER",
-        "_LIBCXXABI_DISABLE_VISIBILITY_ANNOTATIONS", # Only for satic c++abi"
+        # "_LIBCXXABI_DISABLE_VISIBILITY_ANNOTATIONS", # Only for satic c++abi"
```
dzbarsky: nit: "satic"

toolchain/runtimes/BUILD.bazel
```diff
@@ -33,6 +33,45 @@ cc_args_list(
     visibility = ["//toolchain:__subpackages__"],
 )
 
+#TODO(cerisier): Maybe with a _gt_x select we could reuse this everywhere.
```
dzbarsky, line 36: yeah that would be really nice, i think we'll go crazy otherwise if/when we add stage2

runtimes/musl/BUILD.bazel
dzbarsky: oh does it propagate down from the static_lib/dyanmic_lib transition?

cerisier: Yes

toolchain/runtimes/BUILD.bazel
cerisier: I didn't have to add them just yet but as soon as I do, I'll implement this

runtimes/BUILD.bazel
dzbarsky: i think it makes sense to keep just 1 copy here

toolchain/features/runtime_library_search_directories/BUILD.bazel
```diff
@@ -0,0 +1,49 @@
+# This workaround bazel eponym runtime_library_search_directories feature using
```
dzbarsky, line 1: lets file an issue in rules_cc before we forget? :)

dzbarsky (approved):

PR #185, opened 2026-01-10
Implement modulemap and layering_check

dzbarsky:
Closes https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/78

### DIFF
--- a/toolchain/selects.bzl
+++ b/toolchain/selects.bzl
@@ -18,13 +18,19 @@ def platform_extra_binary(binary):
         "@toolchains_llvm_bootstrapped//platforms/config:windows_x86_64": "@prebuilts-extras-toolchain-artifacts-windows-amd64//:%s" % binary,
     })
 
-def platform_cc_tool_map(exec_os, exec_cpu):
+def _tool_repo(exec_os, exec_cpu):
     if exec_os == "macos":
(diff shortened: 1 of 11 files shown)


toolchain/selects.bzl
```diff
@@ -18,13 +18,19 @@ def platform_extra_binary(binary):
         "@toolchains_llvm_bootstrapped//platforms/config:windows_x86_64": "@prebuilts-extras-toolchain-artifacts-windows-amd64//:%s" % binary,
     })
 
-def platform_cc_tool_map(exec_os, exec_cpu):
+def _tool_repo(exec_os, exec_cpu):
     if exec_os == "macos":
```
cerisier, line 22: for another PR, but since we now support macos x86 as exec, this becomes useless ?

cerisier (approved):

toolchain/selects.bzl
dzbarsky: only with from-source llvm bootstrap right?

```
SUPPORTED_EXECS = [
    ("macos", "aarch64"),
    ("linux", "x86_64"),
    ("linux", "aarch64"),
    ("windows", "x86_64"),
    ("windows", "aarch64"),
]
```

we can certainly ship some x86 mac binaries, but IMHO ~nobody is using those for development anymore unless they are accidentally stuck in rosetta x86 hell? M1 shipped 5 years ago :)

PR #189, opened 2026-01-14
Support header parsing feature (--process_headers_in_dependencies)

dzbarsky:
Activate with `bazel build //... --process_headers_in_dependencies`.

This finds a bunch of errors in LLVM :)


toolchain/llvm/llvm.bzl
```diff
@@ -13,10 +16,68 @@ def declare_llvm_targets(*, suffix = ""):
...
+        out = "header_parser.cc",
+        substitutions = {
+            "{CLANG_EXEC_PATH}": "$(execpath :bin/clang++)",
+        },
+        data = [
+            ":bin/clang++",
+        ],
+    )
+
+    cc_runtime_stage0_binary(
```
cerisier: Can't it be just another transition against a `parse_header=False" ?

dzbarsky: I _think_ I tried this and unfortunately it was still cycle, because even if header parsing is not used _for this binary_ the latent ability is still there in the toolchain via the tool map. But maybe i'm misunderstanding uour suggestion, do you want to try?

PR #215, opened 2026-01-24
Add GHA workflow to release prebuilt extras

dzbarsky:


.github/workflows/prebuilts-extras.sh
```diff
@@ -0,0 +1,61 @@
...
+    exit 1
+  fi
+
+  cp "${src}" "${dest}"
+done
+
+(cd release && shasum -a 256 *.tar.zst > SHA256.txt)
+
+if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
+  echo "version=${EXTRAS_VERSION}" >> "${GITHUB_OUTPUT}"
```
cerisier: this is never declared ?

cerisier: I used date YYYYMMDD so far.

dzbarsky: Sorry that should be BRANCH_PAYLOAD I think. Will clean

dzbarsky: And then yeah we can use the same naming pattern

dzbarsky: yerp updated

PR #235, opened 2026-02-02
[prebuilt extras] Fix prebuilt compilation and release

cerisier:
Remove duplicate script and use the ones introduced by @dzbarsky 
Also pass mbedtls flag and add libarchive module decl
Fix target names to match the ones from all_platforms

### DIFF
--- a/.github/workflows/prebuilts-extras.sh
+++ b/.github/workflows/prebuilts-extras.sh
@@ -21,6 +21,7 @@ bazel \
   build \
   --remote_header="x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" \
   --@libarchive//:use_mbedtls=true \
+  --config=bootstrap \
   --config=prebuilt \
   --config=remote \
   --remote_download_outputs=toplevel \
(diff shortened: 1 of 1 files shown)


.github/workflows/prebuilts-extras.sh
```diff
@@ -20,7 +20,7 @@ bazel \
   --bazelrc=".github/workflows/ci.bazelrc" \
   build \
   --remote_header="x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}" \
-  --config=bootstrap \
```
dzbarsky, line 23: i debated this one when writing it, i figured it is better to use latest and greatest compiler and it's a no-op in common case. but probably doesn't matter too much!

.github/workflows/prebuilts-extras.sh
```diff
@@ -30,11 +30,10 @@ rm -rf release
...
-  macos_arm64      darwin-arm64     macos_arm64
-  windows_amd64    windows-amd64-gnu windows_amd64
-  windows_arm64    windows-arm64-gnu windows_arm64
+  linux_amd64_musl linux-amd64-musl extra_bins
+  linux_arm64_musl linux-arm64-musl extra_bins
+  macos_arm64      darwin-arm64     extra_bins
+  windows_amd64    windows-amd64-gnu extra_bins
 )
 
 for ((i=0; i<${#PLATFORMS[@]}; i+=3)); do
```
dzbarsky: heh I guess you could do `i+=2` here and hardcode target_name given its always the same

MODULE.bazel
```diff
@@ -148,6 +148,7 @@ archive_override(
 )
 
 # pbzx support for libarchive
+bazel_dep(name = "libarchive", version = "3.8.1", dev_dependency = True)
```
dzbarsky, line 150: very preexisting but maybe we should sort prod deps above dep deps? Can do it later

PR #238, opened 2026-02-02
Disable compiling from Windows ARM64 hosts

cerisier:
Bazel has issues running there anyway.
And we also need to fix LZMA support for windows arm64.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,7 +85,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, windows-11-arm]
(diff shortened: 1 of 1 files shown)


zbarsky-openai (approved):

.github/workflows/ci.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, windows-11-arm]
```
zbarsky-openai, line 88: Can you leave name commented out so we don’t forget the right one when reenabling

cerisier: Done

PR #292, opened 2026-02-15
[sanitizers] Add support for nsan aka NumericalStabilitySanitizer

cerisier:


3rd_party/llvm-project/21.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -1517,6 +1519,107 @@ genrule(
...
+NSAN_PREINIT_SOURCES = [
+    "nsan_preinit.cpp",
+]
+
+filegroup(
+    name = "nsan_preinit_sources",
+    srcs = ["lib/nsan/" + f for f in NSAN_PREINIT_SOURCES],
+)
+
+NSAN_HEADERS = [
```
dzbarsky, line 1558: this file is getting a little unwieldy. WDYT we write an http_archive-like repo rule that lets us provide an overlay map, so we can inject individual build files into the subdirs? that would also let us drop the list comprehensions adding the source prefixes...

PR #312, opened 2026-02-18
 Don't pass `--unwindlib=none` on macOS 

fmeum:
Silences these warnings encountered when building Bazel for macOS (on macOS).
```
INFO: From Linking external/zstd-jni+/libzstd-jni.so [for tool]:
clang++: warning: argument unused during compilation: '--unwindlib=none' [-Wunused-command-line-argument]
```

### DIFF
--- a/toolchain/args/BUILD.bazel
+++ b/toolchain/args/BUILD.bazel
@@ -713,10 +713,13 @@ cc_args(
     actions = [
         "@rules_cc//cc/toolchains/actions:link_actions",
     ],
-    args = [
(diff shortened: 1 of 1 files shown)


dzbarsky:
Fabian says this is on macOS aarch64 target, perhaps it's related to unwind coming from the SDK?

cerisier:
Let me have a look quickly. I had a plan to remove those warnings but --unwind is a valid option for dynamic library. I think the real reason it shows is because we have another flag that make it dull for the invocation.

PR #313, opened 2026-02-18
Add the `--macos_minimum_os` version used by Bazel

fmeum:
A very old version that makes it possible to build Bazel without changes: https://github.com/bazelbuild/bazel/blob/cdaaf5e1b095dfb63bfaebeb67ecacb543ac949b/.bazelrc#L26


cerisier (commented):
Nice ! However I just realized that we might not handle the host version do we ?? Or does it translate to macos_minimum_os directly in the fragment ??

PR #325, opened 2026-02-21
[mingw] pthread support

dzbarsky:


e2e/rules_cc/BUILD.bazel
```diff
@@ -422,6 +422,16 @@ cc_test(
     tags = ["no-remote"],
 )
 
+cc_binary(
+    name = "pthread_windows",
+    srcs = ["pthread_windows.cc"],
+    copts = ["-pthread"],
+    linkopts = ["-pthread"],
+    target_compatible_with = [
+        "@platforms//os:windows",
```
fmeum: Isn't this fully generic and could run targeting any platform?

dzbarsky: oh, good point! I think I may have had some windows-specific threading apis there in a previous rev. thanks

dzbarsky: oh, I remember why I did this now. This was intended to force this target to run locally instead of remotely, as we are defaulting all actions to (linux) remote

dzbarsky: since it's just a binary we could put it in the cross-compilation suite. But I think actually we should run it as a test :)

Issue #335, opened 2026-02-23
cc targets fail to build when path mapping is enabled for Cpp actions


jjudd:
When trying out this toolchain in our main monorepo I noticed that enabling path mapping with `--experimental_output_paths=strip` and opting into path mapping for the `CppCompile` action with `--modify_execution_info=CppCompile=+supports-path-mapping` causes cc targets to fail to build.

Based on [The State of Path Mapping](https://github.com/bazelbuild/bazel/discussions/22658) discussion on GitHub,  I'd expect `Cpp` actions to work with path mapping. I haven't run into this issue with other cc toolchains, so I believe it is specific to this one.

I created a minimal repro case here: https://github.com/lucidsoftware/toolchains_llvm_bootstrapped_path_mapping_cppcompile_repro

**Summary of the issue*
When path mapping is enabled with `--experimental_output_paths=strip` and path mapping for `Cpp` actions is enabled as follows:
```
--incompatible_modify_execution_info_additive
--modify_execution_info=CppModuleMap=+supports-path-mapping
--modify_execution_info=CppArchive=+supports-path-mapping
--modify_execution_info=CppCompile=+supports-path-mapping
```

Then building cc targets fails with errors like the following:
```
ERROR: /home/redacted/toolchains_llvm_bootstrapped_path_mapping_cppcompile_repro/hello/BUILD.bazel:3:10: Compiling hello/main.cc failed: (Exit 1): clang++ failed: error executing CppCompile command (from cc_binary rule target //hello:hello) external/llvm++http_archive+llvm-toolchain-minimal-21.1.8-linux-amd64/bin/clang++ -target x86_64-linux-gnu -no-canonical-prefixes -Wno-builtin-macro-redefined '-D__DATE__="redacted"' ... (remaining 48 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
hello/main.cc:1:10: fatal error: 'iostream' file not found
    1 | #include <iostream>
      |          ^~~~~~~~~~
1 error generated.
```

I believe this is an issue specifically with `CppCompile` because if I do not enable path mapping for that action, then the build succeeds. To clarify if I don't use this option, then build succeeds: `--modify_execution_info=CppCompile=+supports-path-mapping`.

dzbarsky:
Thank you everyone for chipping in to make this better, this is amazing to see!

dzbarsky:
[...]

PR #343, opened 2026-02-25
module_map: use `ctx.actions.write` instead of `template_dict`

rrbutani:
work towards: https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/335

---

As Fabian mentioned in #335 path mapping artifacts in `cc_args` is the much trickier bit...

(I wrote some pretty [egregious](https://github.com/bazelbuild/rules_cc/commit/9184f2005b713ecefc50c9aac08563ad23305924) [hacks](https://github.com/rrbutani/toolchains_llvm_bootstrapped/commit/20b69f5042d70728cb2113fac13c268d73c5ac2e) as a workaround for now; with those + this PR compiling w/path mapping appears to work)

---

EDIT: I'd forgotten that I also have these flags in my `.bazelrc`:

```python
# `bazel_lib`'s `run_binary` forms command lines (analysis) using unmapped paths
common --modify_execution_info=RunBinary=-supports-path-mapping
# `libcxx_headers_include_search_directory` cannot be path mapped (json file has
# unmapped paths)... for now disable path mapping all CopyDirectory actions
common --modify_execution_info=CopyToDirectory=-supports-path-mapping
```


runtimes/module_map.bzl
```diff
@@ -24,28 +25,36 @@ def _module_map_impl(ctx):
...
         join_with = "\n",
         map_each = lambda file: _textual_header(file, execroot_prefix = execroot_prefix),
         allow_closure = True,
+        expand_directories = False,
     )
 
-    ctx.actions.expand_template(
-        template = ctx.file._module_map_template,
+    module_map_args.add('}')
+    module_map_args.set_param_file_format("multiline")
```
dzbarsky: Not: can you please move this up to creation so it's clearer that it will format correctly when emitted?

runtimes/module_map.bzl
```diff
@@ -24,28 +25,36 @@ def _module_map_impl(ctx):
...
+
+    write_kwargs = {}
+    if bazel_features.rules.write_action_has_mnemonic:
+        write_kwargs["mnemonic"] = "CppModuleMap"
```
dzbarsky, line 52: This is so you can emit a custom --modify- execution? Want to add a recommendation to the README?

rrbutani: Yes; I picked the [same `mnemonic` that `rules_cc` uses](https://github.com/bazelbuild/rules_cc/blob/43ea010aa6070563dc75eb4b1d4cb732425e6a0e/cc/private/compile/cc_compilation_helper.bzl#L359) so that things "just work" for folks following Fabian's guidance [here](https://github.com/bazelbuild/bazel/discussions/22658).

I can add a snippet to the README echoing that guidance (the bit in https://github.com/bazelbuild/bazel/discussions/22658 about C/C++ path mapping)? (with the extra caveat that it'll require Fabian's new changes, Bazel 9.1+)

fmeum: Since you picked the same mnemonic, we fully align with the generic C++ instructions. I don't think we need to add anything to our readme just yet. 

In a separate PR, we could add docs about our supported features (`layering_check`, path mapping, `parse_headers`, ...)

dzbarsky: Works for me! Thanks

PR #350, opened 2026-02-28
Add support for choosing llvm version

cerisier:
This PR does multiple things:

## 1. Move away from the runtime specific archives 

They do not exist anymore starting from 22.x.
This is actually a good thing because it simplifies the reasoning, we now build everything from llvm-raw.

## 2. Introduce choosing the version of LLVM

For now only between 2 hard coded patch version but I think we should extend this to any minor.patch version of support major versions. Worst case, things break but folks can help sending patches.

I made the decision to handle version specific behaviors inside a single set of build files otherwise, given the size of the BUILD definitions we carry, backporting developement to all versions is just brittle. 
Here we have a single set of BUILD files, we encode the changes with `if VERSION > XX`.

Any version after 21 is accepted as best effort.

## TODO:
- [ ] Have a smoke test all versions of LLVM

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -5,7 +5,7 @@ module(
     compatibility_level = 0,
 )
 
-bazel_dep(name = "bazel_lib", version = "3.2.0")
+bazel_dep(name = "bazel_lib", version = "3.2.2")
 bazel_dep(name = "bazel_skylib", version = "1.8.2")
 bazel_dep(name = "bazel_features", version = "1.42.0")
 bazel_dep(name = "platforms", version = "1.0.0")
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,4 +5,5 @@ gazelle export-ignore
 internal_tools export-ignore
 e2e export-ignore
 examples export-ignore
+prebuilt export-ignore
 


cerisier:
We probably need a test to validate that we don't break previous LLVM versions.

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -436,7 +439,7 @@ cc_library(
         "//conditions:default": [],
     }) + select({
         "@llvm//platforms/config:linux_aarch64": [
-            "lib/builtins/cpu_model/aarch64/fmv/mrs.inc",
+            "lib/builtins/cpu_model/aarch64/fmv/hwcap.inc" if IS_LLVM_22_OR_NEWER else "lib/builtins/cpu_model/aarch64/fmv/mrs.inc",
```
dzbarsky: i wonder if it's clearer to do `if LLVM_MAJOR_VER >= 22` here instead of doing it in words? just thinking out loud

3rd_party/llvm-project/x.x/libc/LICENSE.TXT
dzbarsky, line 1: so, now that we are going to depend on llvm tarball anyway, does it actually make sense to vendor these ones?

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -0,0 +1,370 @@
...
+
+filegroup(
+    name = "libcxx_22_headers_files",
+    srcs = LIBCXX_HEADERS_FILES_22,
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "libcxx_headers_files",
+    srcs = [":libcxx_22_headers_files"] if IS_LLVM_22_OR_NEWER else [":libcxx_21_headers_files"],
```
dzbarsky: you can just do `srcs = LIBCXX_HEADERS_FILES_22 if IS_LLVM_22_OR_NEWER else LIBCXX_HEADERS_FILES_21` and drop the above filegroups right?

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -0,0 +1,370 @@
...
+        "_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER",
+    ],
+    copts = [
+        "-faligned-allocation",
+        "-std=c++23",
+
+        "-fvisibility=hidden",
+        "-fvisibility-inlines-hidden",
+
+        # "-fno-exceptions", # if no exceptions
```
dzbarsky, line 178: oh, this reminds me, we shuld probably set this when doing prebuilts (since we set those copts) :)

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -0,0 +1,370 @@
...
+            "-Wno-unused-value",
+        ],
+        "//conditions:default": [],
+    }),
+    includes = [
+        "include",
+        "src",
+    ],
+    hdrs = [(
+        ":libcxx_22_headers_files" if IS_LLVM_22_OR_NEWER else ":libcxx_21_headers_files"
```
dzbarsky: just use `libcxx_headers_files` here?

extensions/llvm_source.bzl
```diff
@@ -4,30 +4,41 @@ load("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
...
+    "//3rd_party/llvm-project/x.x/patches:llvm-sanitizers-ignorelists.patch",
+    "//3rd_party/llvm-project/x.x/patches:no_frontend_builtin_headers.patch",
+    "//3rd_party/llvm-project/x.x/patches:llvm-bzl-library.patch",
+    "//3rd_party/llvm-project/x.x/patches:llvm-driver-tool-order.patch",
+    "//3rd_party/llvm-project/x.x/patches:llvm-dsymutil-corefoundation.patch",
+    "//3rd_party/llvm-project/x.x/patches:compiler-rt-symbolizer_skip_cxa_atexit.patch",
+]
+
+_LLVM_21_SOURCE_PATCHES = _DEFAULT_SOURCE_PATCHES + [
+    "//3rd_party/llvm-project/21.x/patches:llvm-bazel9.patch",
```
dzbarsky, line 21: oh wow I didn't realize so many of these made it into 22, that's great :)

extensions/llvm_source.bzl
```diff
@@ -4,30 +4,41 @@ load("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
...
+    "//3rd_party/llvm-project/21.x/patches:windows_link_and_genrule.patch",
+    "//3rd_party/llvm-project/21.x/patches:bundle_resources_no_python.patch",
+    "//3rd_party/llvm-project/21.x/patches:no_zlib_genrule.patch",
+    "//3rd_party/llvm-project/21.x/patches:no_rules_python.patch",
+    "//3rd_party/llvm-project/21.x/patches:llvm-overlay-starlark.patch",
+    "//3rd_party/llvm-project/21.x/patches:llvm-windows-stack-size.patch",
```
dzbarsky, line 27: wait `llvm-windows-stack-size.patch` one only merged to LLVM this week, there's no way it's in 22. also I didn't send `bundle_resources_no_python.patch` upstream yet.  can you double check this list?

extensions/llvm_source.bzl
```diff
@@ -82,11 +86,52 @@ def _create_llvm_raw_repo(mctx):
...
 
+def _parse_llvm_major(llvm_version):
+    if not llvm_version:
+        fail("LLVM version must not be empty")
+
+    major_token = llvm_version.split(".", 1)[0]
+    if not major_token:
+        fail("Invalid LLVM version '{}': expected '<major>.<minor>.<patch>'".format(llvm_version))
+
+    for i in range(len(major_token)):
```
dzbarsky: replace this loop with `if not major_token.is_digit()` :) https://bazel.build/rules/lib/core/string#isdigit

extensions/llvm_source.bzl
```diff
@@ -115,42 +160,104 @@ _llvm_subproject_repository = repository_rule(
...
-                strip_prefix = "{name}-{llvm_version}.src".format(
-                    name = name,
-                    llvm_version = LLVM_VERSION,
-                ),
-                urls = ["https://github.com/llvm/llvm-project/releases/download/llvmorg-{llvm_version}/{name}-{llvm_version}.src.tar.xz".format(
-                    name = name,
-                    llvm_version = LLVM_VERSION,
-                )],
-            )
+    for name in ["compiler-rt", "libcxx", "libcxxabi", "libunwind"]:
```
dzbarsky, line 266: hmm do we still need to make this subproject repository thing? Can't we just use `files` to inject them all? https://bazel.build/rules/lib/repo/http#http_archive-files (and then tweak the targets to be like `@llvm-project//compiler-rt`

extensions/llvm_source.bzl
```diff
@@ -115,42 +160,104 @@ _llvm_subproject_repository = repository_rule(
...
+        )
+
+def _get_llvm_version(mctx):
+    llvm_version = DEFAULT_LLVM_VERSION
+    module_selected_version = None
+
+    for mod in mctx.modules:
+        module_versions = [tag.llvm_version for tag in mod.tags.version]
+        if len(module_versions) > 1:
+            fail("Only 1 llvm_source.version(...) tag is allowed per module")
```
dzbarsky, line 279: we probably should relax this (in future). I can see wanting incremental compiler upgrades and/or multiple toolchains

toolchain/bootstrap/declare_toolchains.bzl
```diff
@@ -56,7 +57,7 @@ def declare_tool_map(exec_os, exec_cpu):
         # TODO(zbarsky): Previously we did `load("@llvm-project//:vars.bzl", "LLVM_VERSION_MAJOR")`
```
dzbarsky: remove this comment if we're going back to doing that ;)

MODULE.bazel
```diff
@@ -75,6 +75,8 @@ TOOLCHAIN_EXTRAS_VERSION = "20260216"
 ]
 
 llvm_source = use_extension("//extensions:llvm_source.bzl", "llvm_source")
+llvm_source.index(file = "//:llvm_versions.json")
```
dzbarsky: so just to confirm, the idea here is to allow people to bring their own index? (Otherwise we could hardcode it in the extension)

MODULE.bazel
```diff
@@ -75,6 +75,8 @@ TOOLCHAIN_EXTRAS_VERSION = "20260216"
...
+llvm_source.version(llvm_version = LLVM_VERSION)
```
dzbarsky, line 89: do we need this `llvm_source.version`? I though we'd get the default one anyway?

README.md
```diff
@@ -142,6 +142,35 @@ We use a cross-platform reimplementation of `pkgutil` to unpack SDK packages, wh
...
+
+This module allows choosing a specific LLVM version for both the compiler and the runtimes.
+
+### Configure via `use_extension(...)`
+
+To select another LLVM release (for example `22.1.0`), configure the `llvm_source` extension before `use_repo(...)`:
+
+```starlark
+llvm_source = use_extension("@llvm//extensions:llvm_source.bzl", "llvm_source")
+llvm_source.index(file = "@llvm//:llvm_versions.json")  # Optional: default index.
```
dzbarsky: maybe remove this from the example if it's default, and you give an example override below. just to avoid confusion

llvm_versions.json
```diff
@@ -0,0 +1,66 @@
+{
+  "21.1.0-rc1": {
+    "url": "https://github.com/llvm/llvm-project/releases/download/llvmorg-21.1.0-rc1/llvm-project-21.1.0-rc1.src.tar.xz",
```
dzbarsky: wait this is gonna be a huge regression because everyone needs to clone the .tar.xz now, can we point at github archive instead? like https://github.com/llvm/llvm-project/archive/refs/tags/llvmorg-22.1.0.tar.gz

3rd_party/llvm-project/x.x/libc/LICENSE.TXT
cerisier: I can look at it but I think it's a longer one as those to ensure is robust.

extensions/llvm_source.bzl
cerisier: I want this as well, but it's a bit of a larger change since it will need supporting multiple versions of llvm in the same graph

toolchain/bootstrap/declare_toolchains.bzl
cerisier: I left it because I think that's a problem. Here, this will indeed trigger a eager fetch just because we will register toolchains. I think we need to make our own vars.bzl inside some custom repo that generates this from the version passed in .version(

llvm_versions.json
cerisier: I think we were already fetching the .tar.xz, but also the problem with those archive tar.gz is that they are created on the fly and so the streaming of it is super slow compared to CDN. Also those are 2x as big as the `.tar.xz` version :( 

We should really try to get https://github.com/llvm/llvm-project/issues/164537 bumped. Maybe @keith can help pinging the right folks ?

MODULE.bazel
cerisier: Maybe we shouldn't have a default in the extension ?
I have a follow up PR which selects the prebuilts based on that version, in which case there would be only 1 place to state the llvm version (in the MODULE.bazel).

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
cerisier: Done

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
cerisier: Done

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
cerisier: Done

extensions/llvm_source.bzl
cerisier: This one made it, I fixed the others ! good catch !!

extensions/llvm_source.bzl
cerisier: Done

extensions/llvm_source.bzl
cerisier: Fixed ! Thanks and sorry I missed that.

MODULE.bazel
cerisier: Done, no need indeed.

README.md
cerisier: Done

llvm_versions.json
keith: looks like someone got to it, up to the release manager folks to decide if adding a new set of binaries for that is ok, if they're good with it, it should be a small change

dzbarsky: We only really need source archive recompressed, hopefully that's not too contentious...

extensions/llvm_source.bzl
```diff
@@ -115,42 +205,124 @@ _llvm_subproject_repository = repository_rule(
...
+)
+""")
+
+    rctx.file("version.bzl", """\
+LLVM_VERSION_MAJOR = "{major}"
+LLVM_VERSION_MINOR = "{minor}"
+LLVM_VERSION_PATCH = "{patch}"
+LLVM_VERSION = "{version}"
+
+llvm_vars = {{
```
dzbarsky, line 234: should this be LLVM_VARS as it's also a "constant"?

llvm_versions.json
cerisier: Ok so we resorted to just republish everything repacked https://github.com/hermeticbuild/llvm-redist/releases

tools/BUILD.bazel
```diff
@@ -12,6 +13,18 @@ package(default_visibility = ["//visibility:public"])
     for tool in TOOLCHAIN_BINARIES
 ]
 
+native_binary(
```
dzbarsky: note to self: why are we doing this? Shouldn't folks generarally just use the toolchain?

llvm_versions.json
```diff
@@ -0,0 +1,10 @@
+{
+  "21.1.8": {
+    "url": "https://github.com/hermeticbuild/llvm-redist/releases/download/llvmorg-21.1.8-r1/llvm-project-21.1.8.src.tar.zst",
+    "sha256": "64a773b61d91abec7faab139aeda3b8773f2330abc4d97bd26d1513a4e98cf55"
+  },
+  "22.1.0": {
```
dzbarsky, line 6: you gonna backfill this after?

MODULE.bazel
```diff
@@ -74,8 +74,32 @@ TOOLCHAIN_EXTRAS_VERSION = "20260216"
...
+
+BSDTAR_PREBUILTS_SHA256 = {
+    "darwin-arm64": "48c1bd214aac26487eaf623d17b77ebce4db3249be851a54edcc940d09d50999",
+    "linux-amd64-musl": "fff8f72758a52e60fe82beae64b18e7996467013ffe8bec09173d1ba6b66e490",
+    "linux-arm64-musl": "683468ae45d371e4f392b0e5a524440f6f4507d7da0db60d03ff31f3cf951fc3",
+    "windows-amd64-gnu": "f48c81e1812956adb4906c6f057ca856dd280a455e7867d77800e6d5ef9fc81d",
+    "windows-arm64-gnu": "130d69268b0a387bca387d00663821779dd1915557caf7fcbfd5ded3f41074f3",
+}
+
+[
```
dzbarsky: lets do https://github.com/bazel-contrib/toolchains_llvm/blob/7178e8381589bd33ca7c8146d56d2d42a60aa7cc/MODULE.bazel#L29-L38 instead, that way they won't get downloaded multiple times

extensions/llvm_source.bzl
cerisier: This follows what the original LLVM vars.bzl did ha!

tools/BUILD.bazel
cerisier: just as a helper but i can remove them

llvm_versions.json
cerisier: yes, for now it doesn't even work prior to 21.1.8 because we didn't repack the minimal toolchain distributions.

PR #365, opened 2026-03-04
Pass -resource-dir to compiles

keith:
Previously module map files were per toolchain repo, but semantically
these are per target platform instead. This hoists them out of the
nested repos, and also always directs clang to look at the target
include directory. The latter is required for headers that are platform
specific, of which there are very few in the LLVM releases today (only
2).

### DIFF
--- a/3rd_party/libc/glibc/glibc.BUILD.bazel
+++ b/3rd_party/libc/glibc/glibc.BUILD.bazel
@@ -17,6 +17,7 @@ alias(
 )
 
 HDRS = glob([
+    "bits/**/*.h",
     "csu/**",
     "debug/**/*.h",
     "elf/**/*.h",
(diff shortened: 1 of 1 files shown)


toolchain/llvm/llvm.bzl
```diff
@@ -1,45 +1,23 @@
...
         # Grab whichever version-specific dir is there.
-        path = native.glob(["lib/clang/*"], exclude_directories = 0)[0] + "/include",
+        path = native.glob(["lib/clang/*"], exclude_directories = 0)[0],
         visibility = ["//visibility:public"],
     )
 
     # Convenient exports
     native.exports_files(native.glob(["bin/*"]))
 
-    copied_headers = []
```
dzbarsky, line 22: amazing, this contributed like an extra 1K actions and looked kinda clowny

BUILD.bazel
```diff
@@ -17,6 +18,67 @@ alias(
...
+        "@mingw//:mingw_w64_headers_crt_directory",
+        "@mingw//:mingw_w64_headers_include_directory",
+        "@mingw//:mingw_w64_winpthreads_include_directory",
+    ],
+)
+
+include_path(
+    name = "wasm_target_headers",
+    srcs = [
+        "//:builtin_headers",
```
dzbarsky, line 70: can we switch `//:builtin_headers` target to be a `copy_to_directory` of the @llvm-project filegroup or something like that? Otherwise when cross-building I think we end up downloading the target's LLVM archive just for the headers

keith: so i was thinking about using those directly but i don't think we can if the same targets can be used with multiple compiler versions? since we have to match the builtin headers to the current compiler i think.

dzbarsky: So currently we allow you to select an LLVM version but we only allow a single one, I think we take the max one specified by the tags. But I think when we allow having multiple in the build graph (which we'll likely need at some point) we could generate per-version repos like `@llvm-project-21`, `@llvm-project-22` and `//:builtin_headers` could dispatch on a build flag that the toolchains are also `compatible_with` so they stay paired. IDK maybe there's other ways to set it up, we haven't thought it through fully, but I feel like we'll need to solve the same problem anyway for the runtime libraries

keith: fwiw another issue with this is some of those headers are generated with a c++ tool in the llvm repo, so trying to do that now results in a cycle. we could likely adjust the bazel config upstream to use toolchains for that instead of referencing the binary directly, and then maybe we could build that binary in a stage0 config to avoid that issue

PR #370, opened 2026-03-06
Add libfuzzer targets

isaacparker0:
https://github.com/cerisier/toolchains_llvm_bootstrapped/issues/16

### DIFF
--- a/3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
+++ b/3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
@@ -2514,3 +2514,140 @@ cc_runtime_stage0_shared_library(
     shared_lib_name = "libasan.shared.so",
     visibility = ["//visibility:public"],
 )
+
+###
+### compiler-rt/lib/fuzzer/CMakeLists.txt.
+###
+
+FUZZER_SOURCES = [
+    "FuzzerCrossOver.cpp",
+    "FuzzerDataFlowTrace.cpp",
+    "FuzzerDriver.cpp",
+    "FuzzerExtFunctionsDlsym.cpp",
+    "FuzzerExtFunctionsWeak.cpp",
+    "FuzzerExtFunctionsWindows.cpp",
+    "FuzzerExtraCounters.cpp",
+    "FuzzerExtraCountersDarwin.cpp",
+    "FuzzerExtraCountersWindows.cpp",
+    "FuzzerFork.cpp",
+    "FuzzerIO.cpp",
+    "FuzzerIOPosix.cpp",
+    "FuzzerIOWindows.cpp",
+    "FuzzerLoop.cpp",
+    "FuzzerMerge.cpp",
+    "FuzzerMutate.cpp",
+    "FuzzerSHA1.cpp",
+    "FuzzerTracePC.cpp",
+    "FuzzerUtil.cpp",
+    "FuzzerUtilDarwin.cpp",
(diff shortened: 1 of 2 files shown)


3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -2514,3 +2514,140 @@ cc_runtime_stage0_shared_library(
...
+
+filegroup(
+    name = "fuzzer_headers",
+    srcs = ["lib/fuzzer/" + f for f in FUZZER_HEADERS],
+)
+
+# Unlike sanitizer runtimes, libfuzzer is a C++ library that uses the standard
+# library (cassert, vector, string, etc.). Use a subset of the sanitizer flags
+# that omits -nostdinc++ so that the system or toolchain C++ headers are found.
+FUZZER_CFLAGS = select({
```
isaacparker0, line 2597: Let me know if it'd be better to refactor a bit so we're not duplicating much of `SANITIZER_COMMON_CFLAGS`, e.g. refactoring into a subset of base flags that is shared.

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -2514,3 +2514,140 @@ cc_runtime_stage0_shared_library(
...
+    "-funwind-tables",
+    "-fno-stack-protector",
+    "-fvisibility=hidden",
+    "-fno-lto",
+    "-O3",
+    "-gline-tables-only",
+    "-fno-rtti",
+]
+
+cc_library(
```
isaacparker0, line 2614: `cc_runtime_stage0_static_library` currently rejects duplicate symbols, but libfuzzer is a C++ library with legitimate duplicate symbols from standard C++ header usage. To get around that, I just used `cc_library` directly, but I'm assuming might be undesirable because we lose some flag isolation.

Seems like we probably want to just refactor `cc_runtime_stage0_static_library` to support this case, perhaps a `--allow-duplicates` flag?

fmeum: Could you try this: https://github.com/bazelbuild/rules_cc/blob/ee57062efc3ab22132e28af782c70b6d73b052f5/cc/private/rules_impl/cc_static_library.bzl#L266

I built this into cc_static_library for cases like this

isaacparker0: I tried `features = ["-symbol_check"]` on a `cc_runtime_stage0_static_library` target, the feature propagates to the inner target but the validation still runs. Also tried with global `--features=-symbol_check` and didn't disable validation.

I did a little digging and it seems like perhaps the `symbol_check` feature gating is missing for the rule-based API? Do you have an example of this working I could reference @fmeum?

fmeum: rules_cc would need to be modified by adding the `implies` in https://github.com/bazelbuild/rules_cc/blob/ee57062efc3ab22132e28af782c70b6d73b052f5/cc/private/toolchain/unix_cc_toolchain_config.bzl#L287 to the rules-based toolchain as well as not enabling the action by default.

But I wonder whether this static library would be safe on all platforms - is the linker guaranteed to pick the libfuzzer symbol? I wonder why there's a symbol conflict in the first place unless the static library includes parts of the runtime.

PR #371, opened 2026-03-06
[osx] Fix `cups` exclusion rule

ArchangelX360:


extensions/osx.bzl
```diff
@@ -81,6 +80,8 @@ def _osx_extension_impl(mctx):
         excludes.append("usr/include/tcl*")
     if "Tk" not in frameworks:
         excludes.append("usr/include/tk*")
+    if "PrintCore" not in frameworks and "ApplicationServices" not in frameworks:
```
dzbarsky: i think we only need PrintCore right?

ArchangelX360: Oh oops, fixed 😓

PR #381, opened 2026-03-08
Add llvm-link, llvm-cov, llvm-profdata

dzbarsky:


.bazelrc
```diff
@@ -36,7 +36,7 @@ common:prebuilt  --copt=-fomit-frame-pointer
 common:prebuilt  --dynamic_mode=off
 common:prebuilt  --extra_toolchains=//toolchain:all
 # See https://github.com/llvm/llvm-project/blob/497a6d672263343f58eae5a5e697484f4899fbdc/utils/bazel/llvm-project-overlay/llvm/driver.bzl#L11
-common:prebuilt --@llvm-project//llvm:driver-tools=clang-scan-deps,clang,dsymutil,lld,llvm-ar,llvm-cgdata,llvm-cxxfilt,llvm-debuginfod-find,llvm-dwp,llvm-gsymutil,llvm-ifs,llvm-libtool-darwin,llvm-lipo,llvm-ml,llvm-mt,llvm-nm,llvm-objcopy,llvm-objdump,llvm-rc,llvm-readobj,llvm-size,llvm-symbolizer,sancov
+common:prebuilt --@llvm-project//llvm:driver-tools=clang-scan-deps,clang,dsymutil,lld,llvm-ar,llvm-cgdata,llvm-cov,llvm-cxxfilt,llvm-debuginfod-find,llvm-dwp,llvm-gsymutil,llvm-ifs,llvm-libtool-darwin,llvm-link,llvm-lipo,llvm-ml,llvm-mt,llvm-nm,llvm-objcopy,llvm-objdump,llvm-profdata,llvm-rc,llvm-readobj,llvm-size,llvm-symbolizer,sancov
```
dzbarsky, line 39: i feel like this flag shouldn't be on `prebuilt`, so we can more easily test full build without waiting 15 minutes for LTO link, wdyt

cerisier: Do you mean testing the prebuilt out of CI ?

dzbarsky: Yeah. I was building "@llvm-project//llvm" locally to make sure it actually compiles but that doesn't link the full set of tools unless you also pass --config prebuilt, which is very slow :)

dzbarsky: Oh to clarify I meant should we make this flag always apply

PR #396, opened 2026-03-09
Repalce osx.framework with osx.frameworks, since there are usually multiple passed

keith:
I think this is a bit nicer when you have a bunch you want to include


dzbarsky:
> I think this is a bit nicer when you have a bunch you want to include

Should we just deprecate `osx.framework`? I think it's very unlikely you add just 1, especially since as soon as you add any we take away the default list, so you'll probably add those back in :)

PR #397, opened 2026-03-09
Add -framework Foundation by default on macOS

keith:
The apple_support toolchain, which is the default for building objc on
all apple platforms, adds this framework by default. Without this objc
likely doesn't link since it's expecting it. There are some other
default ones that are lower risk that we can add as needed. There's
really no downside since even binaries that don't link Foundation
directly will load it at runtime because libSystem transitively loads
it.

### DIFF
--- a/toolchain/BUILD.bazel
+++ b/toolchain/BUILD.bazel
@@ -216,6 +216,7 @@ cc_args_list(
     name = "macos_toolchain_args",
     args = [
         "//toolchain/args/macos:macos_minimum_os_flags",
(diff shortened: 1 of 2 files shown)


keith:
reference: https://github.com/bazelbuild/apple_support/blob/2643f0185caeaeb170f8bc694be505173768fede/crosstool/cc_toolchain_config.bzl#L1311-L1355

https://github.com/bazelbuild/apple_support/blob/2643f0185caeaeb170f8bc694be505173768fede/crosstool/cc_toolchain_config.bzl#L1986-L1999

PR #398, opened 2026-03-09
Automatically add usr/lib/*.tbd on macOS

keith:
The other tbd files can represent libraries that binaries need that
already exist on the system. This directory was previously ~1mb and with
this is ~3mbs. Alternatively we could allow users to specify libraries
to include as they can with frameworks, but this size differences makes
it seem like maybe it's ok to just include them all.

The only exception is libraries that point to frameworks, which produce
broken symlinks unless the framework is also included, and bazel fails
on that. For now I think we can just ignore those.

Some examples of ones that you might need are libxml2, and compression

### DIFF
--- a/extensions/osx.bzl
+++ b/extensions/osx.bzl
@@ -18,7 +18,7 @@ def _osx_extension_impl(mctx):
     for module in mctx.modules:
         for frameworks_tag in module.tags.frameworks:
             frameworks.extend(frameworks_tag.names)
-        for _ in module.tags.experimental_include_all_sdk_libs:
+        if len(module.tags.experimental_include_all_sdk_libs) > 0:
             experimental_include_all_sdk_libs = True
 
     if not frameworks:
@@ -148,7 +148,7 @@ _frameworks_tag = tag_class(
 )
 
 _experimental_include_all_sdk_libs_tag = tag_class(
-    doc = "Include all usr/lib/*.tbd from the macOS SDK sysroot instead of only the minimal default set.",
+    doc = "Include most usr/lib/*.tbd from the macOS SDK sysroot instead of only the minimal default set. Some libraries that are symlinks to frameworks are still excluded.",
 )
 
 osx = module_extension(


dzbarsky:
So the size is not very concerning to me, but I've usually excluded these and preferred to use @libxml2, @zlib, etc from BCR. The advantage is that it makes it a little harder to introduce nonhermetic deps that will work on Mac but break on linux. Are there correctness reasons to force these to come from the sysroot instead?

dzbarsky:
Ah, are you working on objc support? That would be really cool :)
Maybe for now we add a flag/tag class, something like `osx.experimental_include_all_sdk_libs()` to make this opt-in? and then we can decide the final shape later

extensions/osx.bzl
```diff
@@ -13,10 +13,13 @@ _DEFAULT_FRAMEWORKS = [
 
 def _osx_extension_impl(mctx):
     frameworks = []
+    experimental_include_all_sdk_libs = False
 
     for module in mctx.modules:
         for frameworks_tag in module.tags.frameworks:
             frameworks.extend(frameworks_tag.names)
+        for _ in module.tags.experimental_include_all_sdk_libs:
```
dzbarsky: Include_all = len(tags) > 0. - any cleaner? Up to you!

extensions/osx.bzl
```diff
@@ -109,10 +147,15 @@ _frameworks_tag = tag_class(
     },
 )
 
+_experimental_include_all_sdk_libs_tag = tag_class(
+    doc = "Include all usr/lib/*.tbd from the macOS SDK sysroot instead of only the minimal default set.",
```
dzbarsky: It's not really all since touw excluded some, not sure how best to phrase though

keith: added a bit more detail

extensions/osx.bzl
keith: sg, but still have to gate on only ever setting it to true in case it hits multiple times we don't want it to get reset to false

dzbarsky: Oh true, I forgot

PR #399, opened 2026-03-10
Pass -fno-addrsig to crt objects

keith:
By default clang creates the LLVM_ADDRSIG section. This section gets
invalidated when using `ld.lld -r`. Then when you go to link it later
you see:

```
ld.lld: warning: bazel-out/k8-opt/bin/external/llvm+/runtimes/crt_objects_directory_linux/crtendS.o: --icf=safe conservatively ignores SHT_LLVM_ADDRSIG [index 4] with sh_link=0 (likely created using objcopy or ld -r)
ld.lld: warning: bazel-out/k8-opt/bin/external/llvm+/runtimes/crt_objects_directory_linux/crtbeginS.o: --icf=safe conservatively ignores SHT_LLVM_ADDRSIG [index 13] with sh_link=0 (likely created using objcopy or ld -r)
```

Now we pass `-fno-addrsig` so this is never created instead and
therefore solves the warning

### DIFF
--- a/runtimes/musl/BUILD.bazel
+++ b/runtimes/musl/BUILD.bazel
@@ -129,7 +129,10 @@ cc_library(
 cc_library(
     name = "musl_rcrt1",
     srcs = ["@musl_libc//:crt/rcrt1.c"],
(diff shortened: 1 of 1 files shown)


cerisier (approved):
Oh nice find !

PR #401, opened 2026-03-11
Expose llvm-toolchain-minimal repos

thii:


toolchain/selects.bzl
```diff
@@ -27,6 +27,9 @@ def _tool_repo(exec_os, exec_cpu):
     cpu_part = "amd64" if exec_cpu == "x86_64" else "arm64"
     return "@llvm-toolchain-minimal-%s-%s-%s//" % (LLVM_VERSION, os_part, cpu_part)
 
+def llvm_binary_for_platform(exec_os, exec_cpu, binary):
```
cerisier: platform_llvm_binary_file ?

I feel all those had difficult names from the get go!

thii: yeah seems this is better (since we have a selectable `platform_llvm_binary` above)

MODULE.bazel
```diff
@@ -26,33 +26,20 @@ LLVM_VERSION = "22.1.7"
...
-        )],
-    )
-    for (target, sha256) in LLVM_TOOLCHAIN_MINIMAL_SHA256.items()
-]
+llvm_toolchain_minimal = use_extension("//extensions:llvm_toolchain_minimal.bzl", "llvm_toolchain_minimal")
+llvm_toolchain_minimal.index(file = "//extensions:llvm_toolchain_minimal_index.json")
+llvm_toolchain_minimal.release(
+    llvm_version = PREBUILT_LLVM_VERSION,
+)
+use_repo(
```
dzbarsky, line 34: should we teach this extension to return the `direct_deps` so it's automated? then we don't need to interpolate PREBUILT_LLVM_VERSION below?

dzbarsky (approved):
we might want to rename the extensions to avoid "minimal". Nowadays we are shipping basically all the LLVM tools and I've seen agents (especially claude) get confused by "minimal" and think things are missing

cerisier:
@thii merging now, let me know if you need a bzl helper again if you want to use a specific tool !

PR #421, opened 2026-03-31
Enable BB remote downloader

dzbarsky:


runtimes/mingw/extension/mingw.bzl
```diff
@@ -7,7 +7,7 @@ def _mingw_extension_impl(module_ctx):
     http_archive(
         name = "mingw",
         urls = ["https://sourceforge.net/projects/mingw-w64/files/mingw-w64/mingw-w64-release/mingw-w64-v13.0.0.tar.bz2"],
```
sluongng: It seems like Sourceforge would detect the Go's HTTP client used in BuildBuddy's implementation of Remote Asset API and returned an HTML page instead of the archive 💀 

I think you might want to add `--experimental_remote_downloader_local_fallback` to .bazelrc here and try again 🤔 

Another option would be to use the mirror urls

```
https://zenlayer.dl.sourceforge.net/project/mingw-w64/mingw-w64/mingw-w64-release/mingw-w64-v13.0.0.tar.bz2?viasf=1
```

note that the `viasf=1` is important to keep.

keith: upstream fix https://github.com/buildbuddy-io/buildbuddy/pull/11763

Issue #29, opened 2025-06-09
Support riscv


cerisier:
`clang -target riscv{32:64}-*-*`

cerisier:
Note that only a chosen baseline of RISCV64 is supported and hardcoded for the time being

PR #432, opened 2026-04-07
Add support for 64 bits RISC-V targets

cerisier:
Contributes to #29 

This PR is ready for review now that prebuilts support RISC-V target since #433.

For now, this only brings support for generic linux riscv64 baseline: `rv64gc` and `lp64d`.
The custom `-plugin-opt` is a dirty fix to https://discourse.llvm.org/t/myterious-soft-float-output-in-lto-cache/70753/17

I wired basic gnu and musl test for rules_rust and cross_compaliation e2e.

### DIFF
--- a/3rd_party/libc/glibc/helpers.bzl
+++ b/3rd_party/libc/glibc/helpers.bzl
@@ -19,16 +19,12 @@ def glibc_includes(cpu):
 
     return [
         "include",
-
         "sysdeps/unix/sysv/linux/{}".format(cpu),
     ] + x86_64_variant + x86_parent + riscv64_variant + [
-
         "sysdeps/{}/nptl".format(cpu),
-
         "sysdeps/unix/sysv/linux/generic",
         "sysdeps/unix/sysv/linux/include",
         "sysdeps/unix/sysv/linux",
-
         "sysdeps/nptl",
         "sysdeps/pthread",
         "sysdeps/unix/sysv",
--- a/3rd_party/libc/glibc/helpers.bzl
+++ b/3rd_party/libc/glibc/helpers.bzl
@@ -1,4 +1,29 @@
 def glibc_includes(cpu):
+    if cpu == "riscv64":
+        return [
+            "include",
+            "sysdeps/unix/sysv/linux/riscv/rv64",
+            "sysdeps/unix/sysv/linux/riscv",
+            "sysdeps/unix/sysv/linux/wordsize-64",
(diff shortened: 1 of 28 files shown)


constraints/libc/libc_versions.bzl
```diff
@@ -21,3 +21,10 @@ GLIBCS = ["gnu.{}".format(glibc) for glibc in GLIBC_VERSIONS]
 LIBCS = ["musl"] + GLIBCS
 
 DEFAULT_LIBC = "gnu.2.28"
+
+DEFAULT_LIBCS = {
+    ("linux", "riscv64"): "gnu.2.33",
```
cerisier, line 35: Add comment that llvm for riscv wants kernel >= 5.10
this is poorman's solution

Add a TODO about how to better choose kernel headers versions

3rd_party/libc/glibc/helpers.bzl
```diff
@@ -1,4 +1,29 @@
...
+        return [
+            "include",
+            "sysdeps/unix/sysv/linux/riscv/rv64",
+            "sysdeps/unix/sysv/linux/riscv",
+            "sysdeps/unix/sysv/linux/wordsize-64",
+            "sysdeps/riscv/rv64/rvd",
+            "sysdeps/riscv/rv64/rvf",
+            "sysdeps/riscv/rv64",
+            "sysdeps/riscv/rvd",
+            "sysdeps/riscv/rvf",
```
cerisier: Need to see if this is ok to hard code or if it should be gated

toolchain/args/BUILD.bazel
```diff
@@ -33,14 +33,44 @@ cc_args(
...
+            # Common in most Linux distros.
+            "-march=rv64gc",
+            "-mabi=lp64d",
+        ],
+        "//conditions:default": [],
+    }),
+)
+
+# See https://discourse.llvm.org/t/myterious-soft-float-output-in-lto-cache/70753/16
+# Target features are currently not passed to the LTO linker for RISC-V.
```
dzbarsky, line 57: o.o how did you find this?

dzbarsky (approved):

toolchain/args/BUILD.bazel
cerisier: On the edge of the window. Right before jumping off.

cerisier: Linker complained that ABI was not the same between several objects...

PR #435, opened 2026-04-12
Patch zlib-ng build to add to system include path.

smcallis:
Zlib is included with angle brackets in e.g. CRC.cpp so we need to put the zlib directory on the system include path for it to be used downstream. Otherwise things like libFuzzer fail to build.

TBH I'm not 100% sure this is the correct fix for this but there doesn't seem to be any other way to get LLVM compiling against the included zlib-ng.


cerisier:
@keith @dzbarsky I remember that we had disabled zlib support for this reason and that at some point we were carrying brackets to quotes patches but that we at some point re-emabled zlib support.

Do you remember why and how did it become ok suddenly ?

keith:
maybe it worked because of `--features=external_include_paths`?

smcallis:
This actually seems fixed with Bazel > 8.5.0 too.

PR #436, opened 2026-04-12
Add empty static_libgcc feature to avoid rules_cc adding it for us.

smcallis:
rules_cc autocreates a legacy static_libgcc feature when static_link_cpp_runtimes is enabled, which adds -static-libgcc and emits warnings. This disables those warnings.

### DIFF
--- a/toolchain/features/legacy/BUILD.bazel
+++ b/toolchain/features/legacy/BUILD.bazel
@@ -51,12 +51,21 @@ cc_feature(
     visibility = ["//visibility:public"],
 )
 
+# Define a no-op static_libgcc feature so rules_cc does not backfill its
+# legacy default `-static-libgcc` flag when static_link_cpp_runtimes is enabled.
+cc_feature(
+    name = "static_libgcc",
+    overrides = "@rules_cc//cc/toolchains/features/legacy:static_libgcc",
+    visibility = ["//visibility:public"],
+)
+
 ###
 
 cc_feature_set(
     name = "all_legacy_builtin_features",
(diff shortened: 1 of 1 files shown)


cerisier:
LGTM
I didn't know that we could override a feature twice !

cc @keith for additional eyes.

FYI, we need to support `llvm-libgcc` at some point so we might have to treat this differently then!

cerisier (approved):

keith (approved):

cerisier:
Circling back to this, I realized we already overrode the dependency on that feature:

```
        # Intentionally omitted since we link runtime libraries ourselves and
        # including this flag just raises a warning.
        # "@rules_cc//cc/toolchains/args/static_libgcc:feature",
```

So unless mistaken, we weren't pulling it. 
Did you have a reproducer before after ?

smcallis:
I'll look and see if I can recreate it.

smcallis:
I believe this was caused by me using bazel < 8.5.0 too, sorry for the false alarms.

cerisier:
np ! Can you send a PR to revert ?

smcallis:
Can you hit revert on this one?  I don't have permissions.

keith:
This won't revert cleanly. I changed everything since. What's the goal? For us now this feature is gone

smcallis:
Just responding to @cerisier 's comment, I can revert manually if you like.

cerisier:
Oh if this is gone from main then no need for anything then

smcallis:
Sounds good!

PR #439, opened 2026-04-13
Add missing bzl_library deps

keith:


prebuilt/extras/BUILD.bazel
```diff
@@ -27,7 +27,6 @@ bzl_library(
     visibility = ["//visibility:public"],
     deps = [
         "//prebuilt:mtree",
-        # TODO(zbarsky): They need a library...
```
keith, line 30: i submitted this fix upstream but also we can just use the nested file which does have a bzl_library already

runtimes/musl/extension/BUILD.bazel
```diff
@@ -6,6 +6,8 @@ bzl_library(
     visibility = ["//visibility:public"],
     deps = [
         "@bazel_features//:features",
+        "@bazel_tools//tools/build_defs/repo:cache.bzl",
```
keith, line 9: we don't use these directly but since we're referencing direct files and not a bzl_library target we inherit all the transitive deps too

cerisier (approved):

PR #440, opened 2026-04-13
Remove unused copy_file target

keith:
This also failed in my testing since this rule requires it output a
single file


keith:
not unused

PR #441, opened 2026-04-13
Add no_legacy_features to toolchains

keith:
By default rules_cc adds a lot of features if you don't define them, and
you don't set `no_legacy_features`. At this point the features in
rules_cc for rules based toolchains, as well as the ones in this repo,
cover everything we need. This makes sure we don't get surprising
behavior from those defaults.

### DIFF
--- a/toolchain/features/BUILD.bazel
+++ b/toolchain/features/BUILD.bazel
@@ -154,6 +154,73 @@ cc_feature(
     feature_name = "parse_headers",
 )
 
+cc_feature(
+    name = "coverage_stub",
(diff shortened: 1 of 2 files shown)


keith:
the only features we're actually dropping here are a lot related to `fdo`. I'm highly skeptical those work as intended today, but they would be easy enough to copy if they did

dzbarsky (approved):

keith:
these are the ones being dropped:

```
fdo_instrument
fdo_optimize
cs_fdo_instrument
cs_fdo_optimize
fdo_prefetch_hints
autofdo
propeller_optimize_thinlto_compile_actions
propeller_optimize
```

PR #442, opened 2026-04-13
Add runtimes_all to some select statements

keith:
Without this doing a `bazel build --nobuild ...` fails


cerisier (approved):
LGTM

I think the //toolchain:runtime API might not be the best because now we have "runtimes_all" that gives nothing for targets where runtimes_stage1 give something so is bit counter intuitive.

Very interested if you have an idea for making this better sementically.

keith:
yea that logic made me assume that complete actually isn't load bearing today? otherwise that is probably wrong?

PR #443, opened 2026-04-13
Fix running gazelle by default

keith:
Otherwise on linux x86_64 this fails with pic issues. We likely need a
better fix


fmeum:
I would have expected https://github.com/bazel-contrib/rules_go/commit/2c439988276d777de8ebffd92b38a0e92e7aa1f0 to fix this, do you understand why that's not the case?

keith:
the logic is a bit different from the commit message:

https://github.com/bazel-contrib/rules_go/commit/2c439988276d777de8ebffd92b38a0e92e7aa1f0#diff-f6aaf6101bc7030a9b8d09d47a1af5918ef80ca113d867cf8b781d8d4d9d5722R479-R485

it's not defaulting it on on all supported platforms, i believe it's only doing the platforms that themselves default to pie?

PR #451, opened 2026-04-14
Fix formatting for new buildifier version

keith:
Notably order of deps can matter for libcxx headers and friends


keith:
notably the new version sorts `includes`, which i have omitted here since we rely on ordering. im going to try and annotate the relevant ones

PR #463, opened 2026-04-15
Pass -Wno-modules-import-nested-redundant for layering_check

keith:
If you have layering_check enabled, all your imports come through the
modulemaps. If you have a file like this:

```
// foo.c
include <stdio.h>

void foo() {
include "foo.def"
}

// foo.def
include <stdio.h>
```

You get this warning:

```
external/+_repo_rules+llvm-project/lldb/source/Plugins/Process/Utility/RegisterInfos_x86_64_with_base_shared.h:10:1: error: redundant #include of module 'crosstool.../../../../../external/llvm++glibc+glibc_headers_x86_64-linux-gnu.2.28/include' appears within function 'GetRegisterInfo_i386' [-Wmodules-import-nested-redundant]
   10 | #include <stdint.h>
      | ^
```

### DIFF
--- a/toolchain/args/BUILD.bazel
+++ b/toolchain/args/BUILD.bazel
@@ -90,6 +90,7 @@ cc_args(
         "-Wno-module-import-in-extern-c",
         # This warning would mean we had an error configuring the toolchain or modulemap
         "-Werror=incomplete-umbrella",
+        "-Wno-modules-import-nested-redundant",
     ],
 )
 


keith:
alternatively we could make this just `Wno-error=` but realistically if things don't expect to be compiled with modulemaps, its likely there would be many useless warnings anyways

cerisier (approved):

Issue #16, opened 2025-06-09
Compile and link against Fuzzer


cerisier:
Adds instrumentation for coverage-guided fuzzing using libFuzzer, automatically generating inputs to maximize code path coverage.

Flag: -fsanitize=fuzzer
Lib: libclang_rt.fuzzer

---

And FuzzerNoMain:

Provides fuzzing instrumentation without linking a default main() function, allowing custom fuzz drivers.
Flag: -fsanitize=fuzzer-no-link
Lib: libclang_rt.fuzzer_no_main

PR #470, opened 2026-04-19
feat: add support for libclang_rt.fuzzer runtime

smcallis:
- Added build rules for the libFuzzer runtime in the compiler-rt overlay.
- Added toolchain flags enabled via //config:fuzzer and exported the runtime into the clang resource directory.
- Compiles fuzz targets with -fsanitize=fuzzer-no-link and links libclang_rt.fuzzer explicitly instead of relying on the driver to add extra runtimes.
- Keeps sanitizer composition explicit: //config:fuzzer can be combined with //config:asan, //config:ubsan, and other sanitizer configs without pulling in UBSan by default.
- Updated runtime exports so multiple sanitizer configs compose cleanly in the resource directory.
- Added E2E coverage for plain libFuzzer plus fuzzer+asan and fuzzer+ubsan compositions.


smcallis:
This would close out #16 I believe

cerisier:
Any reason why not letting the driver pick it up ? In general I think we want to lean towards driver compliance since the future of this project might involve generate sysroots and resourcedirs vs passing everything manually via clang flags.

runtimes/BUILD.bazel
```diff
@@ -174,47 +174,74 @@ copy_to_resource_directory(
     srcs = {
         "//runtimes/compiler-rt:clang_rt.builtins.static": "libclang_rt.builtins",
     } | select({
+        "//config:fuzzer_enabled": {
+            "//runtimes/compiler-rt:clang_rt.fuzzer.static": "libclang_rt.fuzzer",
+            "//runtimes/compiler-rt:clang_rt.fuzzer_no_main.static": "libclang_rt.fuzzer_no_main",
+            "//runtimes/compiler-rt:clang_rt.fuzzer_interceptors.static": "libclang_rt.fuzzer_interceptors",
+        },
+        "//conditions:default": {},
```
cerisier, line 182: Here you add those to the resource-dir which is great but why not let the driver pick them up ?

smcallis: I was pattern matching on the other runtimes, I assumed they needed to be included to link properly.  Is that not so?

cerisier: For all sanitizers and runtimes we actually let the driver pick them up and avoid adding them manually as linker inputs because most often than not, the driver handles link put order in a specific way (that's the case for all sanitizers). Also we want to defer as much work to the driver as possible.

cerisier: By that I mean that we just put libraries in the resource-dir with the correct name and let -fruntime trigger the correct driver include behavior.

smcallis: Happy to do it however you think's best, maybe a quick summary of what code changes I should make?

cerisier:
I updated the PR. Basically, every `-fsanitizer=` makes the clang driver auto pick up some libs as linker inputs in the resource-dir.

Since you already copied the right files with the right names in the resource-dir for fuzzer, i just changed the linkopt to `-fsanitizer=fuzzer` which will make the driver auto pick up fuzzer.a and ubsan.a in the resource-dir.

cerisier:
Last comment would be about usage of py vs sh since that would diverge from other tests but maybe it's required ?

smcallis:
Yeah I ran into trouble getting sh to work properly on windows but can't
remember why now, I can try to port them to .sh instead.

cerisier:
Are you windows based ? if so you can be very handy haha

smcallis:
No I'm not windows based that's what makes it hard =D. I see you got it
though, awesome!

PR #471, opened 2026-04-19
feat: add support for libclang_rt.xray runtime

smcallis:
- Added build rules for the XRay and xray-basic runtimes in the compiler-rt overlay.
- Added toolchain flags enabled via //config:xray and exported the runtimes into the clang resource directory.
- Reset XRay in the runtime and stage0 transitions so compiler-rt does not recurse back into an XRay-enabled toolchain while building the XRay runtime itself.
- Removed the redundant sanitizers_interface_headers dependency from sanitizer_common and sanitizer_common_libc. Those targets already carry the include roots they use, do not include the public sanitizer headers, and the extra dependency could trigger include-tree collisions in the stage0/runtime build path.
- Added an E2E that builds an instrumented binary and verifies xray-basic writes a non-empty trace file.


smcallis:
An example trace output from a simple mandelbrot binary which can be loaded in [speedscope](https://speedscope.app) et al.  Fun =D
[mandelbrot.xray.json.gz](https://github.com/user-attachments/files/26874887/mandelbrot.xray.json.gz)

toolchain/args/BUILD.bazel
```diff
@@ -677,6 +677,56 @@ cc_args_list(
...
+    name = "xray_linker_flags",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:link_actions",
+    ],
+    # The main toolchain does not add -resource-dir on macOS, so XRay needs
+    # it explicitly to find the runtime under the hermetic resource dir.
+    # Passing the same resource dir twice on other platforms is harmless.
+    args = [
+        "-resource-dir",
+        "{resource_dir}",
```
cerisier: Should we just not do this for now so that all behavior remains the same on macOS ? I can work on adding this back by supporting resource-dir on macOS entirely, or happy to get a PR if you have cycles for that :)

smcallis: Removed, I'll see if I have time to add macOS resource-dir support.

cerisier:
Fixed and merged. Thanks !

smcallis:
Awesome!

PR #481, opened 2026-04-23
Revert "Temporarily disable coverage features"

keith:
Reverts hermeticbuild/hermetic-llvm#479


keith:
creating so i don't forget, i'll add a test here too

PR #482, opened 2026-04-23
Revert "Temporarily disable coverage features (#479)"

keith:
This reverts commit 9023b08197f35e3b2675b1eb6ff178f949bc888b.


keith:
creating so i don't forget, i'll add a test here too

PR #486, opened 2026-04-23
Remove prefer_pic_for_opt_binaries by default

keith:
This is off in the default toolchain


cerisier:
Sadly I added that because of lack of control on non pic object generation. I think we discussed this briefly on slack somewhere lol

PR #490, opened 2026-04-24
Use bazel's variables for pic/pie logic

keith:
This makes sure we pass `-fno-PIC` if pic is disabled, since llvm
defaults to it enabled for linux, but rules_go defaults it to disabled.
This also removes our custom constraints for this and respects the
variables from rules_cc. rules_cc uses `pic` for compiling and
`force_pic` for linking, them being on or off is enough for our cases.


keith:
this fixes building rules_go host tools where this mismatched what rules_go expected

keith:
ok but this breaks a different case, will look

constraints/pie/BUILD.bazel
```diff
@@ -1,16 +0,0 @@
-package(default_visibility = ["//visibility:public"])
-
-constraint_setting(
```
dzbarsky, line 3: keep in mind these are used in https://github.com/hermeticbuild/rules_rs/blob/2fec8588c806720ddfe9a78e52dac4fe43020548/rs/experimental/platforms/triples.bzl#L21 so we need a replacement for that as well!

keith: yea, ci is covering this, looking for a working version.

keith: ideally it would be using bazel's stuff for that

keith: if it is possible, but that's what i'll figure out

dzbarsky: cool good thing we added rules_rs to CI here 👍

PR #491, opened 2026-04-24
Use zig macOS sysroot by default

keith:
This avoids downloading from the unstable dmg url for the simplest cases
that don't require real frameworks or extra libraries.


dzbarsky:
We have a stable archive now :)

PR #492, opened 2026-04-24
Fix cgo pic issues

keith:
The binaries of clang we ship default to pic on linux, but most other
tools don't expect that and default to it off. This results in pic link
failures with golang. Now we pass `-no-pie` to force that off for
executables unless bazel's pic is enabled.


toolchain/args/linux/BUILD.bazel
```diff
@@ -15,6 +15,20 @@ cc_args(
...
 )
 
+cc_args(
+    name = "non_pie_executable_link_flags",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:link_executable_actions",
+    ],
+    # Clang 22 defaults Linux links to PIE. Preserve Bazel's non-force_pic
+    # executable semantics for GNU targets while still letting force_pic's
+    # later -pie flag win.
```
fmeum: Instead of having this come first and be overridden, could this be a `cc_args` that is applied when `force_pic` is not enabled?

keith: yea done

fmeum (approved):
@dzbarsky

toolchain/args/linux/BUILD.bazel
```diff
@@ -15,6 +15,20 @@ cc_args(
...
+    actions = [
+        "@rules_cc//cc/toolchains/actions:link_executable_actions",
+    ],
+    # Clang 22 defaults Linux links to PIE. Preserve Bazel's non-force_pic
+    # executable semantics for GNU targets.
+    args = select({
+        "//platforms/config:gnu": ["-no-pie"],
+        "//conditions:default": [],
+    }),
+    requires_none = "@rules_cc//cc/toolchains/variables:force_pic",
```
dzbarsky, line 29: i think when I first looked at this I missed this `requires_none` bit (or maybe it's new in the latest revision?) Anyway I feel like this makes sense, @cerisier do you agree?

dzbarsky (approved):

cerisier:
I think I agree with this choice. Also because I have grown to become anti driver. I'm doing a final memory refresh on clang driver logic for PIC/PIE decision and will report!

Thanks for thinking of this, that's clever !

PR #514, opened 2026-04-29
Add llvm-xray to multicall binary.

smcallis:


smcallis:
ARM errors still seem internal.

cerisier:
Is it upstreamable to llvm-project ?

smcallis:
I don't see why not, I can try to do that too if you like.

cerisier:
Yeah we usually try to upstream all the patches as ideally we don't have to carry any ! Of course we'll carry them as long as they are not merged.

cerisier:
Any reason why you need this as part of multicall ? If the target is available in llvm-project, you can use it as well. We use the multicall target for the prebuilt binaries used in actions of this toolchain generally. 

@dzbarsky can arbitrate on that more than me

smcallis:
Let me try using the llvm-project one and I'll report back.

dzbarsky:
> Any reason why you need this as part of multicall ? If the target is available in llvm-project, you can use it as well. We use the multicall target for the prebuilt binaries used in actions of this toolchain generally.
> 
> @dzbarsky can arbitrate on that more than me

Generally I am ok with shipping additional tools in the multicall (though we do risk making final binary link even more painful than it is now). However, the underlying tools needs to use the table-driven Opts library for CLI parsing, not the `cl::Opt` one. The problem with cl::Opt is that it defines the flags as globals, so they pollute help text/registration for unrelated commands.

[llvm-xray is using cl::Opt ](https://github.com/llvm/llvm-project/blob/6a825893a464c1489f7a70960215a2b6b421fc57/llvm/tools/llvm-xray/xray-account.cpp#L35-L48)so we should not include it here, at least not without porting to Opt library. I submitted a few such ports to upstream and they are extremely painful to get past review so I wouldn't hold my breath there :) Probably best to build from source if you can, we reexport @llvm-project so all targets should be accessible

smcallis:
Sounds reasonable, I can use @llvm-project llvm-xray binary just fine, just lots more files to compile the first time but I suppose that's what remote caches are for =D

PR #516, opened 2026-04-29
Fix linking failure of libunwind on Linux targets

### LINKED ISSUE #513: `undefined symbol: _Unwind_GetIP` when compiling an `.so`
I'm using `--@llvm//config:experimental_stub_libgcc_s`, and the produced `.so` is having some undefined symbols which makes the library not loadable at runtime:
```
$ llvm-nm -m -u libmylib.so | grep _Unwind_GetIP
                 U _Unwind_GetIP
                 U _Unwind_GetIPInfo
```

When using `--@llvm//config:experimental_stub_libgcc_s`, we link against an stubbed `libgcc_s`, an empty `.a`.
Maybe I'm facing something related to https://discourse.llvm.org/t/as-needed-breaks-our-build-with-undefined-symbols/75505

Any idea?

ArchangelX360:
Closes #513

### DIFF
--- a/runtimes/libunwind/BUILD.bazel
+++ b/runtimes/libunwind/BUILD.bazel
@@ -32,11 +32,16 @@ config_setting(
 
 copy_to_directory(
     name = "libunwind_library_search_directory",
-    srcs = [
-        ":unwind",
-    ] + select({
+    srcs = select({
+        "@platforms//os:linux": [":libunwind.static"],
+        "//conditions:default": [":unwind"],
+    }) + select({
         ":stub_libgcc_s": [":gcc_s"],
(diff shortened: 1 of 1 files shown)


cerisier:
I commented in the issue and i'm not sure this is the right fix in the end :(

ArchangelX360:
No problem, I'll check these comments tomorrow and comment there ☺️

dzbarsky:
Discussed/debugged over slack, this was a combination of bug in rules_rust not adding transitive shared libs to runfiles + rpath, and user code using `filegroup` to pick out the shared library without merging in the runfiles :)

PR #533, opened 2026-05-03
Add qemu smoke tests for non mainstream cross compilation

cerisier:
We could also just qemu all the way.

### DIFF
--- a/e2e/cross_compilation/test_hello_qemu.sh
+++ b/e2e/cross_compilation/test_hello_qemu.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+resolve_runfile() {
+    local path="$1"
+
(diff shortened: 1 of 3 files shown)


e2e/cross_compilation/test_hello_qemu.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BIN="$BINARY"
```
dzbarsky, line 4: I thought with rootpath you can just use the binary as passed. This bit smells like codex doing runfiles hacks that aren't needed :)

dzbarsky (approved):

e2e/cross_compilation/test_hello_qemu.sh
cerisier: I reused the sanitizer shell script without paying much attention lol

PR #536, opened 2026-05-04
Introduce a dedicated constraint for linux kernel version

cerisier:
This allows users to choose whatever they need.
Also this will allow setting conveniance paltforms for popular distros with tuple glibc + kernel.

The constraint is major.minor only, nested under `kernel/linux` so `freebsd` and others will have it's own constraint.
I kept the old mapping where kernel was selected based on glibc when unconstrained.


kernel/extension/make_select_kernel_headers_repository_target.bzl
```diff
@@ -1,22 +1,80 @@
...
-            apparent_target = "@linux_kernel_headers_{}.{}//:{}".format(arch_to_kernel_arch(target_arch), kernel_version, bazel_target)
+            apparent_target = _kernel_headers_repository_target(target_arch, kernel_version, bazel_target)
             selection["@llvm//platforms/config:{}_{}_{}".format(target_os, target_arch, libc_version)] = apparent_target
 
     return select(selection)
+
+def _make_select_kernel_headers_repository_target_from_linux_kernel(bazel_target):
+    """Select explicit Linux kernel constraints, falling back to the libc-derived default."""
+    selection = {
+        "@llvm//constraints/kernel/linux:{}".format(kernel_version): ":{}".format(_version_alias_name(kernel_version, bazel_target))
```
dzbarsky: nit: you can drop the `":{}".format(` bit here and below and just use the target name without the colon :)

dzbarsky (approved):

constraints/kernel/linux/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
+load("@bazel_lib//:bzl_library.bzl", "bzl_library")
+load(":declare_linux_kernel_constraints.bzl", "declare_linux_kernel_constraints")
+
+package(default_visibility = ["//visibility:public"])
+
+constraint_setting(
+    name = "version",
+    default_constraint_value = "unconstrained",
```
fmeum, line 8: If there is no meaningful default, I think it's cleaner to just not have a default? Not entirely sure though.

constraints/kernel/linux/declare_linux_kernel_constraints.bzl
```diff
@@ -0,0 +1,8 @@
+load(":linux_kernel_versions.bzl", "LINUX_KERNEL_VERSIONS")
+
+def declare_linux_kernel_constraints():
+    for version in LINUX_KERNEL_VERSIONS:
```
fmeum: Inlining this as a list comprehension would make the BUILD file easier to read

fmeum (approved):

constraints/kernel/linux/BUILD.bazel
cerisier: I can do this but it would be breaking change because today, kernel version is taken from a glibc correspondance map. 

If I set a default version here, that will break the correspondance for all users who are not aware of this constraint.

dzbarsky: We might want to swap this when we add the autodetect-host-platform repo rules to @platforms_contrib? People already are a bit confused why they need to specify a host glibc constraint, having to specify a host kernel constraint before there's machinery to make it painless might get annoying

constraints/kernel/linux/declare_linux_kernel_constraints.bzl
cerisier: Agree

constraints/kernel/linux/BUILD.bazel
dzbarsky: And then for this toolchain specifically, I think defaulting to effectively the oldest one we support should be a safe conservative thing to do, right? Still beats non-bazel glibc builds defaulting to current :)

constraints/kernel/linux/declare_linux_kernel_constraints.bzl
cerisier: Done

PR #541, opened 2026-05-06
Inject our extra build files as an additional overlay for LLVM source repo

dzbarsky:
The current per-project external repos lead to issues on Windows hosts, see https://github.com/hermeticbuild/hermetic-launcher/actions/runs/25379788722/job/74605335934?pr=21


cerisier (approved):
This looks much cleaner.

However, the drawback is that it will fetch llvm everytime we operate on the build files no ? Or at least unarchive llvm at each iteration?

dzbarsky:
> This looks much cleaner.
> 
> However, the drawback is that it will fetch llvm everytime we operate on the build files no ? Or at least unarchive llvm at each iteration?

I don't think it will. `files` operates by symlinking them into place, and it does not `rctx.watch` them (except on Windows hosts).

PR #544, opened 2026-05-07
Mark libgcc_s flag as universal scope

dzbarsky:
Otherwise it won't propagate to exec and then rust builds don't work because processwrapper doesn't build :)


cerisier:
This is bazel 9 only no ??

dzbarsky:
> This is bazel 9 only no ??

The attribute is in skylib so I think it parses fine? But gets ignored in early bazel

PR #547, opened 2026-05-08
Add Nativelink as a User

MarcusSorealheis:
The main usability factor here for us was that builds becomes easier to debug because failures show up as Bazel platform/toolchain mismatches instead of vague host setup problems.

Thank you to the whole team here, immensely!


cerisier:
You're very welcome ! Thank you for giving it a chance, and do not hesitate to file issues if you encounter any, we're committed to making this VERY robust.

cerisier (approved):

PR #554, opened 2026-05-10
Add osx.libraries tag to osx extension

ArchangelX360:
to extend macOS sysroot with specific usr/lib/*.tbd libraries without pulling the full SDK

Example at JetBrains:
```
osx.libraries(names = ["libz"])  # needed by GraalVM native-image link
```

### DIFF
--- a/extensions/osx.bzl
+++ b/extensions/osx.bzl
@@ -35,12 +35,15 @@ def _get_from_archive(mctx):
 
 def _osx_extension_impl(mctx):
     frameworks = []
+    libraries = []
     experimental_include_all_sdk_libs = False
     from_archive = _get_from_archive(mctx)
 
     for module in mctx.modules:
         for frameworks_tag in module.tags.frameworks:
             frameworks.extend(frameworks_tag.names)
+        for libraries_tag in module.tags.libraries:
+            libraries.extend(libraries_tag.names)
         if len(module.tags.experimental_include_all_sdk_libs) > 0:
             experimental_include_all_sdk_libs = True
 
@@ -75,6 +78,9 @@ def _osx_extension_impl(mctx):
             "usr/lib/libpthread.tbd",
             "usr/lib/libSystem*",
         ])
+
+    for library in libraries:
+        includes.append("usr/lib/%s*" % library)
 
     for framework in frameworks:
         includes.append("System/Library/Frameworks/%s.framework/*" % framework)
@@ -185,6 +191,12 @@ _frameworks_tag = tag_class(
     },
 )
 
+_libraries_tag = tag_class(
+    attrs = {
+        "names": attr.string_list(mandatory = True),
+    },
+)
+
 _experimental_include_all_sdk_libs_tag = tag_class(
     doc = "Include most usr/lib/*.tbd from the macOS SDK sysroot instead of only the minimal default set. Some libraries that are symlinks to frameworks are still excluded.",
 )
@@ -204,6 +216,7 @@ osx = module_extension(
     tag_classes = {
         "from_archive": _from_archive_tag,
         "frameworks": _frameworks_tag,
+        "libraries": _libraries_tag,
         "experimental_include_all_sdk_libs": _experimental_include_all_sdk_libs_tag,
     },
 )


cerisier (approved):

ArchangelX360:
We're holding the merge on that, I need to validate that it's needed for the use-case of JetBrains. We're unsure with @dzbarsky, it seems depending on `zlib` BCR woudl be enough, but I have to try it.

ArchangelX360:
Actually thinking about it a bit more, I think I'm convinced we need this `libraries` in the extension.

The suggestion was to link against `zlib` BCR module instead, but design-wise I find it a bit weird to ask users to link against some BCR module when their runtime will run against macOS's one.
Sure, most of these libraries probably have a stable API, but still it looks like a mitigation.

Also, there might cases also where some macOS SDK lib will have no BCR equivalent, and that would be unfortunate to have to build one just because we don't support exposing these libraries like we do for frameworks.

Finally, I suppose that it would require macOS targets to build `zlib` module, while the alternative is given a stub file to the linker.

WDYT @dzbarsky?

dzbarsky:
> Actually thinking about it a bit more, I think I'm convinced we need this `libraries` in the extension.
> 
> The suggestion was to link against `zlib` BCR module instead, but design-wise I find it a bit weird to ask users to link against some BCR module when their runtime will run against macOS's one. Sure, most of these libraries probably have a stable API, but still it looks like a mitigation.
> 
> Also, there might cases also where some macOS SDK lib will have no BCR equivalent, and that would be unfortunate to have to build one just because we don't support exposing these libraries like we do for frameworks.
> 
> Finally, I suppose that it would require macOS targets to build `zlib` module, while the alternative is given a stub file to the linker.
> 
> WDYT @dzbarsky?

chatted offline - I suppose the complexity is pretty low so I don't mind merging this. I still think if you are actually using zlib you'd be better off using zlib-ng and statically linking, but sure, you may have some precompiled bits that weren't statically linked and you don't want to fix it up :)

asymingt:
Was this ever tested on a GitHub runner? I've been using it to compile opencv locally with a darwin-sandbox on Tahoe and it works really well. However the moment I try to run the same Bazel command in CI with the setup-bazel action LLVM throws "framework is not supported" errors I've tried macos-26 and macos -15 workers. Here's an example failire: https://github.com/intrinsic-opensource/ros-central-registry/actions/runs/28733583290/job/85203612548#step:7:1

Here's the workflow:https://github.com/intrinsic-opensource/ros-central-registry/blob/lyrical%2F2026-06-08/.github%2Fworkflows%2Fcheck_distribution.yaml

The .bazelrc and MODULE.bazel files are both generated by CI with this script: https://github.com/intrinsic-opensource/ros-central-registry/blob/lyrical%2F2026-06-08/tools%2Fci%2Fsetup_workspace.py.

dzbarsky:
> Was this ever tested on a GitHub runner? I've been using it to compile opencv locally with a darwin-sandbox on Tahoe and it works really well. However the moment I try to run the same Bazel command in CI with the setup-bazel action LLVM throws "framework is not supported" errors I've tried macos-26 and macos -15 workers. Here's an example failire: https://github.com/intrinsic-opensource/ros-central-registry/actions/runs/28733583290/job/85203612548#step:7:1
> 
> Here's the workflow:https://github.com/intrinsic-opensource/ros-central-registry/blob/lyrical%2F2026-06-08/.github%2Fworkflows%2Fcheck_distribution.yaml
> 
> The .bazelrc and MODULE.bazel files are both generated by CI with this script: https://github.com/intrinsic-opensource/ros-central-registry/blob/lyrical%2F2026-06-08/tools%2Fci%2Fsetup_workspace.py.

I think that error might be triggered when you have a dangling symlink to a framework that didn't get included, I would check the contents of the AVfoundation.

Btw while looking I noticed that https://github.com/intrinsic-opensource/ros-central-registry/blob/b9f36f681e0d682115f1b4efc465d45e8eed242d/tools/ci/setup_workspace.py#L141 changes action_env but it should be repo_env. Also setup-bazel action shouldn't be necessary, runners already come with bazelisk preinstalled :)

asymingt:
> I think that error might be triggered when you have a dangling symlink to a framework that didn't get included, I would check the contents of the AVfoundation.

That's a good point. I tried to resolve all such links and include an exhaustive list. At [one point](https://github.com/intrinsic-opensource/ros-central-registry/blob/25beb7afbd3a904e74319a542144cbd144efffdf/tools/ci/setup_workspace.py#L407-L447) this list was extremely long. In the end `Cocoa` needed `AppKit`, which in turn needed `ApplicationServices` which needed system-level CUPS headers, which are presumably sandboxed out of the build.

> 
> Btw while looking I noticed that https://github.com/intrinsic-opensource/ros-central-registry/blob/b9f36f681e0d682115f1b4efc465d45e8eed242d/tools/ci/setup_workspace.py#L141 changes action_env but it should be repo_env. 

Thank you! Fixed in https://github.com/intrinsic-opensource/ros-central-registry/commit/58ba32c8c0e77a8f79a6efb10c5040d2ea881e69

> Also setup-bazel action shouldn't be necessary, runners already come with bazelisk preinstalled :)

That's news to me. Since I am overwriting a lot of my ROS modules all the time I've been comforted by the `repository-cache: false`  option provided by `setup-bazel` (eventually I'll switch it to true when the first official release lands).

dzbarsky:
> > I think that error might be triggered when you have a dangling symlink to a framework that didn't get included, I would check the contents of the AVfoundation.
> 
> That's a good point. I tried to resolve all such links and include an exhaustive list. At [one point](https://github.com/intrinsic-opensource/ros-central-registry/blob/25beb7afbd3a904e74319a542144cbd144efffdf/tools/ci/setup_workspace.py#L407-L447) this list was extremely long. In the end `Cocoa` needed `AppKit`, which in turn needed `ApplicationServices` which needed system-level CUPS headers, which are presumably sandboxed out of the build.

Are you sure it wasn't looking for the sysroot CUPS headers, [which we only include if PrintCore is enabled?](https://github.com/hermeticbuild/hermetic-llvm/blob/8b16ee1379fb4de532064c3c23af0c9cf004d3c2/extensions/osx.bzl#L155-L156). (Granted that logic is just a minor optimization because we thought CUPS was legacy and only used by PrintCore, but if that's not true, we can just remove those 2 lines and always include them in the default sysroot, or add ApplicationServices to that condition to include CUPS. Feel free to give that a shot!)

asymingt:
With a little bit of extra work I managed to isolate and add the extra frameworks that were needed to get a clean build. I guess the sharp edge here is not being able to add the top-level frameworks and have the transitive dependency frameworks pulled in automatically for you. I needed to iterate and fix for 20 or so cycles to ultimately land on the right set for my application.

dzbarsky:
> With a little bit of extra work I managed to isolate and add the extra frameworks that were needed to get a clean build. I guess the sharp edge here is not being able to add the top-level frameworks and have the transitive dependency frameworks pulled in automatically for you. I needed to iterate and fix for 20 or so cycles to ultimately land on the right set for my application.

yeah, agreed that it's annoying, I've also hit this sharp edge before :/ I wonder if we can examine the extracted files for dangling symlinks or something. OTOH it's possible that some framework's header includes another framework, but you end up not including that header, so you don't need it. (I'm also not sure if it's possible to examine symlink status in starlark). Open to other ideas to make this better!

PR #558, opened 2026-05-12
Add support for libstdc++

cerisier:
This PR brings, from-source compilation of libstdc++.so and the ability to compile and link programs against it as part of the build.

## Usage

For now, we still use `static_runtime_lib` and `dynamic_runtime_lib` and let Bazel choose the right implementation.
Because of this, it means that libstdc++ is only usable with:
```
cc_binary(
    ...
    linkstatic=False,
)
```

Forcing both libstdc++ and unwind lib (there should only be one per process) to be linked dynamically.
It is not ideal but moving out of this would likely need going away from the `static_runtime_lib` and `dynamic_runtime_lib` `cc_toolchain` attributes.

## Implementation

Part of the changes required for this PR has been:
- introducing an indirection on the cxxstdlib chosen both for headers and artifact.
- introducing a selection mechanism for this (for now constraint but I see the limitations)
- introducing an additional runtime stage to allow C only toolchains (for tools required to build libstdc++)

### Some decisions about libstdc++ itself:
- I chose to keep pulling and building latest version and not offer version selection like the glibc.
- Linux GNU only -- Libstdc++ handles a wide variety of hosts but this is an ok start. We can add more later.
- Static libstdc++ is not supported even tho it should be possible without too much efforts.

### Autoconf Implementation:
Biggest party pooper is autoconf. And libstdc++ makes HEAVY usage of many autoconf features.

For this I had to:
- Introduce a generic-like lightweight private autoconf probe system that allows for switching to an autoconf ruleset later.
- Introduced a sync system to ensure we account for and use the same checks performed by the gcc/libstdc++ source.
- Implemented all project specific .m4 check definitions in files that mimic the names of their original source for easier migration
- Implemented `configure.ac.bzl` analog translated Starlark file which uses all checks for the supported configuration (linux gnu). Easy migration as well.

For all config knobs, I've separated between:
- target-derived - hardcoded by platform (like atomic policy)
- policy knobs (to be turned later in build_settings)
- prob knobs (result of checks)

I didn't want to introduce build settings just yet as we're not sure about the future of the constraint model for cxxstdlib and also because we absolutely need to support `exec` vs `target` build settings (`host_` prefix).

### Build files

Headers are the most involved here because of heavily generated files from preprocessing of `config.h` which is generated by our autoconf implementation.

The other ones are:
- `c++config`
- `gthr`
- `largefile-config.h`

Which all uses `config.h` as input and extract / transform to produce a new header.

For all those I chose to remain verbatim with libstdc++ which uses shell scripts. This means that this is not portable on non unixy windows.

For the artifact, as always, I respect naming, structure and conventions of the source Makefiles definitions.

### Toolchain wiring

Libstdc++ is wired the same way as libc++ with the exception that the autoconf probes requires a hosted C toolchain (with a libc).

For this, I introduced a new runtime stage to be able to wire libc and kernel headers.

For now, only glibc is supported but as soon as we support `musl.so` there is no reason why we couldn't support libstdc++ + musl.

### Notes on Autoconf

We can either keep this lightweight in-tree implementation or switch to one of:
- rules_autoconf
- rules_cc_autoconf
- starconf

All three have different approaches and some are already quite complex projects.

A part from checks, missing required features for us are:
- Ability to add c/c++ flags, c only and c++ only flags, link flags to a check (Filed https://github.com/periareon/rules_cc_autoconf/issues/135 for rules_cc_autoconf)
- Ability to add custom inputs to a check (include check on a user provided header for examples)
- Composed probes (check who depends on other check and inherits it's output definitions)

## TODO

- [x] README mention of libstdc++


runtimes/cxxstdlib/BUILD.bazel
```diff
@@ -0,0 +1,116 @@
...
+    cmd = "echo 'libstdc++ is only supported as a dynamic C++ runtime; static libstdc++ is intentionally unsupported.' >&2; exit 1",
+)
+
+genrule(
+    name = "libstdcxx_windows_unsupported",
+    outs = ["unsupported/windows-libstdcxx"],
+    # TODO(corentin): replace this with MinGW libstdc++ DLL/import-library
+    # wiring once Windows libstdc++ support is intentionally added.
+    cmd = "echo 'libstdc++ is currently supported only for Linux targets; Windows/MinGW libstdc++ is intentionally unsupported.' >&2; exit 1",
+)
```
cerisier, line 82: Maybe we can do something different than genrule but since it will fail anyway mabe we don't care.

3rd_party/gcc/gcc.BUILD.bazel
```diff
@@ -0,0 +1,1171 @@
...
+        "libstdc++-v3/include/c_global/cuchar",
+        "libstdc++-v3/include/c_global/cwchar",
+        "libstdc++-v3/include/c_global/cwctype",
+    ],
+)
+
+filegroup(
+    name = "libstdcxx_c_compatibility_headers",
+    srcs = [
+        "libstdc++-v3/include/c_compatibility/complex.h",
```
dzbarsky, line 245: would injecting `files = {"libstdc++-v3/BUILD.bazel": "libstd.BUILD.bazel"}` allow us to clean the prefixes on all of these files? I guess it would make `local_repository` more annoying?

3rd_party/gcc/gcc.BUILD.bazel
```diff
@@ -0,0 +1,1171 @@
...
+)
+
+# GCC installs libgcc/unwind-generic.h as unwind.h for libsupc++ consumers.
+copy_file(
+    name = "libsupcxx_unwind_h",
+    src = "libgcc/unwind-generic.h",
+    out = "libgcc/unwind.h",
+)
+
+# MinGW does not support the alias used by the generic tinfo.cc path here, so
```
dzbarsky, line 482: ho ho ho

3rd_party/gcc/gcc.BUILD.bazel
```diff
@@ -0,0 +1,1171 @@
...
+    ]),
+)
+
+# Generated source adapters mirror the configure-selected source substitutions
+# referenced from libstdc++-v3/src/c++98/Makefile.am and
+# libstdc++-v3/src/c++11/Makefile.am.
+copy_file(
+    name = "libstdcxx_cxx98_cxxlocale_cc",
+    src = libstdcxx_c_locale_cc(),
+    out = "libstdcxx_generated/c++98/c++locale.cc",
```
dzbarsky, line 566: should all of these by `allow_symlink = True` to avoid an actual copy action?

3rd_party/gcc/gcc.BUILD.bazel
```diff
@@ -0,0 +1,1171 @@
...
+            # TODO: enable print.cc once MinGW native_handle_type selection is modeled.
+        ],
+        "//conditions:default": [":libstdcxx_cxx23_print"],
+    }),
+)
+
+# The shared library version and linker flags should stay aligned with GCC's
+# libstdc++-v3/acinclude.m4 shared library and symbol-version settings.
+cc_runtime_stage1_shared_library(
+    name = "libstdc++.so.6.0.35",
```
dzbarsky: maybe leave a note here to update this when bumping URL and vice-versa? make it easier for the bots

constraints/cxxstdlib/cxxstdlib_versions.bzl
```diff
@@ -0,0 +1,6 @@
+CXXSTDLIBS = [
+    "libcxx",
```
dzbarsky, line 2: should we call it libcxx or libc++? ditto with the other?

runtimes/cxxstdlib/BUILD.bazel
dzbarsky: eh :)

runtimes/cxxstdlib/BUILD.bazel
```diff
@@ -0,0 +1,116 @@
...
+        "//constraints/cxxstdlib:libstdcxx",
+        "@platforms//os:windows",
+    ],
+)
+
+alias(
+    name = "headers_include_search_directory",
+    actual = select({
+        ":libstdcxx_linux": "//runtimes/libstdcxx:headers_include_search_directory",
+        ":libstdcxx_windows": ":libstdcxx_windows_unsupported",
```
dzbarsky, line 42: i thought you had some mingw expand_template stuff above? Is that aspiriational?

runtimes/cxxstdlib/BUILD.bazel
```diff
@@ -0,0 +1,116 @@
...
+    outs = ["unsupported/windows-libstdcxx"],
+    # TODO(corentin): replace this with MinGW libstdc++ DLL/import-library
+    # wiring once Windows libstdc++ support is intentionally added.
+    cmd = "echo 'libstdc++ is currently supported only for Linux targets; Windows/MinGW libstdc++ is intentionally unsupported.' >&2; exit 1",
+)
+
+filegroup(
+    name = "static_runtime_lib",
+    srcs = select({
+        ":libstdcxx_linux": [":libstdcxx_static_runtime_lib"],
```
dzbarsky, line 87: can you remind me, how does unwinding work? It is already baked in?

runtimes/libstdcxx/BUILD.bazel
```diff
@@ -0,0 +1,467 @@
...
+
+filegroup(
+    name = "config_h_header",
+    srcs = [":config_h"],
+    output_group = "config_h",
+)
+
+# Configured public headers mirror libstdc++-v3/include/Makefile.am. The
+# selected source labels come from configure.host fields in :configure.bzl.
+copy_file(
```
dzbarsky, line 61: allow_symlink

runtimes/libstdcxx/BUILD.bazel
```diff
@@ -0,0 +1,467 @@
...
+        "}",
+    ],
+)
+
+cc_binary(
+    name = "libstdcxx_toolchain_dynamic_link_smoke",
+    srcs = [":libstdcxx_link_smoke_src"],
+    linkstatic = False,
+    target_compatible_with = [
+        "@platforms//os:linux",
```
dzbarsky: can you use that config setting that combines them? I guess it was private?

runtimes/libstdcxx/config_define_status.txt
```diff
@@ -0,0 +1,214 @@
...
+_GLIBCXX_USE_SYSCTL_HW_NCPU unsupported
+_GLIBCXX_USE_TMPNAM modeled
+_GLIBCXX_USE_UCHAR_C8RTOMB_MBRTOC8_CXX20 modeled
+_GLIBCXX_USE_UCHAR_C8RTOMB_MBRTOC8_FCHAR8_T modeled
+_GLIBCXX_USE_UTIME modeled
+_GLIBCXX_USE_UTIMENSAT modeled
+_GLIBCXX_USE_WCHAR_T modeled
+_GLIBCXX_USE_WIN32_SLEEP modeled
+_GLIBCXX_USE__GET_OSFHANDLE unsupported
+_GLIBCXX_VERBOSE build-setting-later
```
dzbarsky: do we need this checked in so we don't forget? Or it's just a working copy?

runtimes/libstdcxx/config_probe.bzl
```diff
@@ -0,0 +1,537 @@
...
+    )
+    return source
+
+def _declare_compile_probe(ctx, cc_toolchain, check, template, source_placeholder, output_placeholder):
+    source = _declare_source(ctx, check)
+    stem = ctx.attr.name + "_" + check.name.lower()
+    result = ctx.actions.declare_file(stem + ".result")
+    log = ctx.actions.declare_file(stem + ".log")
+
+    ctx.actions.run_shell(
```
dzbarsky: ugh :) ideally we'd have a wrapper binary instead of bash runner, but this is fine for now, maybe just leave a TODO

runtimes/libstdcxx/config_probe.bzl
```diff
@@ -0,0 +1,537 @@
...
+        inputs = depset(
+            direct = [source],
+            transitive = [cc_toolchain.all_files],
```
dzbarsky: do all_files need to be both inputs and tools? I thought tools get automatically merged?

runtimes/libstdcxx/config_probe.bzl
```diff
@@ -0,0 +1,537 @@
...
+        ),
+        outputs = [result, log],
+        tools = cc_toolchain.all_files,
+        arguments = [
+            template.tool,
+            source.path,
```
dzbarsky: using `.path` means path mapping won't actually work, swap to `ctx.actions.args()`?

runtimes/libstdcxx/config_probe.bzl
```diff
@@ -0,0 +1,537 @@
...
+    echo "  \\"error_constants_dir\\": \\"$error_constants_dir\\","
+    echo "  \\"symver_style\\": \\"$symver_style\\","
+    echo "  \\"checks\\": {"
+    cat "$checks_json"
+    echo
+    echo "  }"
+    echo "}"
+} > "$summary"
+""",
+        execution_requirements = {"supports-path-mapping": "1"},
```
dzbarsky: ditto, path mapping won't work until `.path` is fixed

runtimes/libstdcxx/headers.bzl
```diff
@@ -0,0 +1,233 @@
...
+        command = """set -eu
+out="$1"
+config_h="$2"
+{
+    grep 'define _DARWIN_USE_64_BIT_INODE' "$config_h" || true
+    grep 'define _FILE_OFFSET_BITS' "$config_h" || true
+    grep 'define _LARGE_FILES' "$config_h" || true
+} > "$out"
+""",
+        execution_requirements = {"supports-path-mapping": "1"},
```
dzbarsky: ask our bot to fix them all :) Maybe we should discourage these patterns in agents.md?

toolchain/runtimes/BUILD.bazel
```diff
@@ -36,60 +36,58 @@ cc_args_list(
...
 
 cc_args_list(
-    name = "toolchain_args",
+    name = "stage0_toolchain_args",
     args = [
         # Common default compile flags.
         "//toolchain/args:target_flags",
         "//toolchain/args:no_absolute_paths_for_builtins",
         "//toolchain/args:deterministic_compile_flags",
+        # "//toolchain/args:module_map",
```
dzbarsky: ?

runtimes/cxxstdlib/BUILD.bazel
cerisier: It is not baked into `libstdc++.so.6`. For the libstdc++ runtime selection, the dynamic runtime set still carries `//runtimes/libunwind:libunwind.shared` alongside `:libstdcxx_dynamic_runtime_lib`, so the unwinder remains a separate runtime artifact. The libstdc++ shared object links its own libsupc++ pieces, but unwinding is still provided through the explicit libunwind runtime wiring.

runtimes/libstdcxx/config_define_status.txt
cerisier: Yes, it is intended to be checked in. It moved under `runtimes/libstdcxx/docs/config_define_status.txt`, and `//runtimes/libstdcxx/tests:config_define_audit_test` uses it as the reviewed status ledger for generated/ported config defines. That gives us a stable file to update when GCC changes instead of treating the audit output as a disposable working copy.

runtimes/libstdcxx/autoconf/cc_configure_probe.bzl
```diff
@@ -0,0 +1,373 @@
+# Generic CC compile/link probe execution for autoconf-style checks.
+# Keep this file free of runtime policy, generated header, and config.h
+# emission details; callers own those decisions.
+
+load("@rules_cc//cc:action_names.bzl", "ACTION_NAMES")
+load("@rules_cc//cc:find_cc_toolchain.bzl", "CC_TOOLCHAIN_TYPE", "find_cc_toolchain", "use_cc_toolchain")
+load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
+
+_SANITIZER_SETTINGS = [
```
dzbarsky: we should pull this into a common .bzl, since we'll need to reuse these for the other transitions as well

runtimes/libstdcxx/tests/BUILD.bazel
```diff
@@ -0,0 +1,177 @@
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
+load("@rules_shell//shell:sh_test.bzl", "sh_test")
+
+package(default_visibility = ["//visibility:public"])
+
+_AUTOCONF_AUDIT_DATA = [
```
dzbarsky, line 8: hahaha i like your setup :)

toolchain/BUILD.bazel
```diff
@@ -17,6 +17,7 @@ string_setting(
     values = [
         "stage0",  # depend on compiler-only: CRTs, builtins, libc, static libraries.
         "stage1",  # depends on above: libunwind
+        "stage1_hosted",  # stage1 plus target C headers for probes and small build tools.
```
dzbarsky, line 20: i'm a bit meh on the name but don't have a better suggestion :/

README.md
```diff
@@ -161,6 +161,51 @@ Behind the scenes, code is compiled with headers for the selected glibc and link
...
+
+```starlark
+cc_binary(
+    name = "app",
+    srcs = ["main.cc"],
+    linkstatic = False,
+)
+```
+
+Without `linkstatic = False`, Bazel prefers the static C++ runtime path. For
```
dzbarsky: are you sure this isn't just because of our .bazelrc flag? I thought by default it would still link dynamic_runtime_library even for cc_library? not sure though

dzbarsky (approved):

toolchain/BUILD.bazel
cerisier: I can always call it stage2 and shift everything. Those are not important names, only the semantic is.

README.md
cerisier: Let me double verify

runtimes/libstdcxx/autoconf/cc_configure_probe.bzl
cerisier: Oh I think I reverted a change were I reused configure_builder_for_runtimes ! LEt me check

README.md
cerisier: Double checked this. It is not only our `.bazelrc` flag.

With Bazel's default `--dynamic_mode=default`, `cc_binary.linkstatic` defaults to `True`, so a libstdc++ binary without `linkstatic = False` selects the toolchain `static_runtime_lib` path. I verified that with:

```sh
bazel build --config remote --platforms=:linux_x86_64_libstdcxx //:main_default
```

That fails on `@llvm//runtimes/cxxstdlib:libstdcxx_static_runtime_lib`, as expected.

The explicit dynamic target works:

```sh
bazel build --config remote --platforms=:linux_x86_64_libstdcxx //:libstdcxx_main_dynamic
```

and `cquery` shows it goes through `@llvm//runtimes:dynamic_runtime_lib` -> `@llvm//runtimes/cxxstdlib:libstdcxx_dynamic_runtime_lib`.

The `.bazelrc`/prebuilt nuance is also real though: `--dynamic_mode=off` forces the static runtime path even with `linkstatic = False`, so it cannot be combined with this libstdc++ mode. I updated the README wording to spell out both parts.

PR #561, opened 2026-05-13
Introduce 3-stage toolchain bootstrap build graph for FDO-optimized releases

dzbarsky:


.bazelrc
```diff
@@ -60,7 +60,6 @@ common:rust    --@llvm//config:experimental_stub_libgcc_s=True
 
 common --bes_results_url=https://app.buildbuddy.io/invocation/
 common --bes_backend=grpcs://remote.buildbuddy.io
-common --experimental_remote_downloader=grpcs://remote.buildbuddy.io
```
cerisier, line 57: Why ?

toolchain/BUILD.bazel
cerisier: I think you're right and let's called that bootstrap_stage, let's also have some comments on the various stage values to explain what they are and what they are used for.

toolchain/bootstrap/bootstrap_binary.bzl
```diff
@@ -1,71 +1,48 @@
...
-
         # we are compiling final programs, so we want all runtimes.
         "//toolchain:runtime_stage": "complete",
-
-        # We want to build those binaries using the prebuilt compiler toolchain
         "//toolchain:source": "prebuilt",
-        "@llvm-project//llvm:driver-tools": _LLVM_TOOLS,
+        "//command_line_option:compilation_mode": "opt" if needs_llvm_optimization else settings["//command_line_option:compilation_mode"],
+        "//command_line_option:copt": _append_unique(copts, _LLVM_TOOL_COPTS) if needs_llvm_optimization else copts,
+        "//command_line_option:features": features + ["thin_lto"],
```
cerisier: is this normal ?

toolchain/bootstrap/declare_toolchains.bzl
```diff
@@ -15,17 +15,40 @@ def _validate_static_library_tool(prefix):
...
         ],
     )
 
+def _bootstrap_tool_binary(name, actual, **kwargs):
+    bootstrap_binary(
+        name = name,
+        actual = actual,
+        visibility = ["//visibility:public"],
+        **kwargs
+    )
```
cerisier: only for default visibility ?

cerisier (commented):
Finally took the time to really understand what's going on here.

We have a 3-stage bootstrap.
- prebuilt to compile a stage1 from source to benefit from arbitrary LLVM not just prebuilt
- stage1 to compile a instrumented from source 
- instrumented compiles zstd to get a profile
- <QUESTION> instrumented compiles the final compiler

Unless mistaken, it feels wrong to use the instrumented compiler to compile the final compiler if we don't get profiles out of it. In which case it would always be slower than using the stage1 no ?

Otherwise, I understand enough to merge, even tho I think it complexifies quite a lot the bootstraph phase! I think what cmplexlifies things is that there is no support in rules_cc for getting profile data out per compile action which is why we move out of toolchain resolution and rules_cc for those steps.

toolchain/bootstrap/declare_toolchains.bzl
dzbarsky: i'll just remove this, we can put viz in bootstrap_tool_kwargs

toolchain/bootstrap/bootstrap_binary.bzl
dzbarsky: I think so? it's how you enable a feature via transition. We want to build an LTOed binary so this is the best way

cerisier: I mean, this means that all bootstrap tool binary are built with thinlto ? how were we doing before ?

.bazelrc
dzbarsky: it was causing issues again with some archive, i'll bring it back and check

dzbarsky:
> Finally took the time to really understand what's going on here.
> 
> We have a 3-stage bootstrap.
> 
>     * prebuilt to compile a stage1 from source to benefit from arbitrary LLVM not just prebuilt
> 
>     * stage1 to compile a instrumented from source
> 
>     * instrumented compiles zstd to get a profile
> 
>     * instrumented compiles the final compiler
> 
> 
> Unless mistaken, it feels wrong to use the instrumented compiler to compile the final compiler if we don't get profiles out of it. In which case it would always be slower than using the stage1 no ?

I think my reasoning was that the instrumented one was LTO'ed while the stage1 wasn't, so maybe it will make final compile faster? But you're right that the instrumentation will slow it down so it's unclear if it's worth it, and it's cleaner conceptually to reuse the stage1. I'll rearrange it.

> 
> Otherwise, I understand enough to merge, even tho I think it complexifies quite a lot the bootstraph phase! I think what cmplexlifies things is that there is no support in rules_cc for getting profile data out per compile action which is why we move out of toolchain resolution and rules_cc for those steps.

Yeah, once this merge let me put up an alternative that builds an LLVM as the profiled workload and we can decide how we feel :D

toolchain/bootstrap/bootstrap_binary.bzl
dzbarsky: changed to only the ones that matter

PR #564, opened 2026-05-17
Support the full matrix of runtimes as static and dynamic

cerisier:
## Summary

This PR moves runtime selection out of `cc_toolchain.static_runtime_lib` / `dynamic_runtime_lib` and into rule-based toolchain features.

Runtime libraries and related link inputs are now attached to `cc_args(data = ...)` gated by `static_linking_mode` / `dynamic_linking_mode`. This makes the
selected link mode determine which runtime artifacts enter the link action, instead of relying on opaque `cc_toolchain` runtime attributes.

## Driver-Based Runtime Linking

This also shifts more responsibility back to the compiler driver.

Instead of having rules_cc force runtime linker inputs directly, the toolchain now provides the driver with the right sysroot and library search inputs, then lets   the driver expand the standard runtime `-l` behavior where possible. As part of that, this removes more `-nostd*` flags and only keeps explicit linker flags where the driver cannot infer them reliably.

## Runtime Coverage

This PR also adds missing static or dynamic counterparts for existing runtimes, most notably:

- static `libstdc++`
- dynamic `musl`
- matching static/dynamic handling for runtimes that were previously one-sided

glibc remains dynamic-only for now because it is still represented through the existing stub path.

## Behavior Changes

A key behavior change is that binaries and tests can now select different runtime inputs in all supported configurations.

For example, a `cc_binary` may be linked statically while a `cc_test` is linked dynamically, so they may compile, link, and load different runtime artifacts.
Previously this behavior mostly applied to `libc++` and `libunwind`; now it applies consistently across the runtime set, except for glibc, which remains unconditionally dynamic.

## Runtime Discovery

This PR does not configure runtime rpaths.

Dynamically linked outputs must rely on the loader’s normal search behavior or external runtime setup. This change is only about correct link-time input and argument selection, not runtime library discovery.

### DIFF
--- a/runtimes/libcxx/BUILD.bazel
+++ b/runtimes/libcxx/BUILD.bazel
@@ -1,5 +1,5 @@
+load("@bazel_lib//lib:copy_file.bzl", "copy_file")
 load("@bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
-load("//runtimes:defs.bzl", "stub_library")
 
 alias(
     name = "libcxx.static",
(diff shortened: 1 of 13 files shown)


runtimes/libcxx/BUILD.bazel
```diff
@@ -25,20 +25,75 @@ alias(
...
-stub_library(
-    name = "c++abi",
+copy_file(
+    name = "libcxxabi_static_archive",
+    src = ":libcxxabi.static",
+    out = "static/libc++abi.a",
+    allow_symlink = True,
+)
+
+genrule(
```
dzbarsky: Write_file?

cerisier: Fixed

PR #577, opened 2026-06-01
Support macOS dSYM generation

cerisier:
## Summary
- add macOS rules_cc dSYM generation via hermetic llvm-dsymutil
- route macOS link actions through a C link-wrapper that expands params, preserves dsym output metadata, and runs clang, dsymutil, then optional llvm-strip
- normalize wrapper argv handling around temp response files, following apple_support's flow shape while keeping only the needed behavior


toolchain/llvm/llvm.bzl
```diff
@@ -177,6 +230,20 @@ def declare_llvm_targets(*, suffix = ""):
         ],
     )
 
+    cc_tool(
+        name = "link-wrapper",
+        src = "@llvm//tools/internal:link-wrapper",
+        data = [
+            ":clangxx_file",
```
dzbarsky, line 237: why do we define these as a separate filegroup instead of sticking to the bin + suffix pattern the rest of these use?

cerisier: I guess it used the existing shape for other binaries ! let me see.

toolchain/exec_stage0_binary.bzl
```diff
@@ -7,3 +7,9 @@ _exec_stage0_binary_builder = with_cfg(
 )
 
 exec_stage0_binary, _exec_stage0_binary_internal = configure_builder_for_runtimes(_exec_stage0_binary_builder, "stage0").build()
+
+_exec_stage1_hosted_binary_builder = with_cfg(
```
dzbarsky: should this .bzl file be renamed or these moved?

tools/internal/link_wrapper.c
```diff
@@ -0,0 +1,496 @@
...
+#ifdef _WIN32
+#include <process.h>
+#include <windows.h>
+#else
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+#endif
+
+#define STRIP_DEBUG_SYMBOLS_ARG "LLVM_STRIP_DEBUG_SYMBOLS"
```
dzbarsky, line 15: STRIP_DEBUG_SYMBOLS_ARG -> STRIP_DEBUG_SYMBOLS_ENV?

toolchain/llvm/llvm.bzl
cerisier: cxxfilt_file and llvm_nm_file were done like this so it used the same shape

toolchain/exec_stage0_binary.bzl
cerisier: indeed.

tools/internal/link_wrapper.c
cerisier: It's not an env sadly :/ because of lack of this. https://github.com/bazelbuild/rules_cc/issues/728

dzbarsky: or maybe not :)

tools/internal/BUILD.bazel
```diff
@@ -11,3 +12,10 @@ cc_binary(
     srcs = ["static_library_validator.cc"],
     visibility = ["//visibility:public"],
 )
+
+exec_stage1_hosted_binary(
```
dzbarsky: i didn't realize we already had ability to put tools in this extra stage, should we do same for header-parser and static lib validator?

cerisier: So those 2 are in c++. I want to allow this in another pass by unifying the toolchains completely and allow c++ then. Right now we have runtime toolchains that do not yet support c++.

toolchain/features/legacy/BUILD.bazel
```diff
@@ -53,6 +54,43 @@ cc_feature(
...
 
+# General case when not targetting macOS or when targetting macOS but not generating a dSYM.
+# nested_args so that we can do requires_not_none on strip_debug_symbols and requires_none on dsym_path.
+cc_nested_args(
+    name = "strip_debug_symbols_when_no_dsym",
+    args = ["-Wl,-S"],
+    requires_none = "@rules_cc//cc/toolchains/variables:dsym_path",
+)
+
+# Marker argv for the linker_wrapper when targetting macOS only.
```
keith:
```suggestion
# Marker argv for the linker_wrapper when targeting macOS only.
```

keith (approved):

PR #584, opened 2026-06-03
Use rules_cc env requires_not_none for sentinel strip marker

cerisier:
Blocked by https://github.com/bazelbuild/rules_cc/pull/731


cerisier:
Was already done in #638

PR #586, opened 2026-06-03
Implement `external_include_paths`

martin4861:
Implement `external_include_paths` which was added to the rule-based toolchain in https://github.com/bazelbuild/rules_cc/pull/323.

Setting `--features=external_include_paths` alters the `-I` to `-isystem` and helps to silence warnings from external headers.

Unsure how to implement a proper integration test.

### DIFF
--- a/toolchain/cc_toolchain.bzl
+++ b/toolchain/cc_toolchain.bzl
@@ -12,6 +12,7 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
             "@llvm//toolchain/features:archive_param_file",
             "@llvm//toolchain/features:prefer_pic_for_opt_binaries",
             "@llvm//toolchain/features:parse_headers",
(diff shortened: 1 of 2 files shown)


keith (approved):
lgtm, note this has to be before normal include path options until https://github.com/bazelbuild/rules_cc/issues/729 is fixed, looks to me like it is here

carlosgalvezp:
> this has to be before normal include path options

No, the order does not matter. Normal include paths must be fully removed for this to take effect on Clang (it works on GCC).

See example where the `-isystem` line comes first:

```
$ clang++ -c -Wall -Werror -isystem include -iquote include main.cpp
In file included from main.cpp:1:
include/foo.h:3:7: error: unused variable 'x' [-Werror,-Wunused-variable]
    3 |   int x;
      |       ^
1 error generated.
```

PR #589, opened 2026-06-03
Use new toolchains support for macOS min OS

keith:
We can now use toolchain (no not those toolchains) variables in cc_args
for this instead of a giant select()


dzbarsky (approved):

toolchain/args/macos/BUILD.bazel
```diff
@@ -25,37 +24,18 @@ cc_sysroot(
 
 #TODO(cerisier): Fix with new cc_args BuildSettingInfo formatter once it is ready.
 #                See https://github.com/hermeticbuild/hermetic-llvm/issues/302
-macos_minimum_os_flag(
```
dzbarsky, line 28: Snip TODO above this

PR #598, opened 2026-06-06
Allow building LLVM with ABI-breaking checks

dzbarsky:
Add the default-off `@llvm-project//llvm:enable_abi_breaking_checks` build setting to LLVM's Bazel overlay. Enabling the setting generates `LLVM_ENABLE_ABI_BREAKING_CHECKS=1`; it does not enable ordinary assertions or change `NDEBUG`.

The downstream source patch corresponds to llvm/llvm-project#203739. The patch runs after the LLVM 21 and LLVM 22 version-specific patches because those patches also modify `utils/bazel/llvm-project-overlay/llvm/BUILD.bazel`.

Validation:

- Applied `llvm-abi-breaking-checks.patch` to cached post-version-patch LLVM 21 and LLVM 22 overlays.
- Verified the generated overlays contain `enable_abi_breaking_checks`, `abi_breaking_checks_enabled`, and `LLVM_ENABLE_ABI_BREAKING_CHECKS=1` with no `-UNDEBUG` or assertion-setting names.
- Ran Buildifier on `extensions/llvm_source.bzl`.

### DIFF
--- a/3rd_party/llvm-project/x.x/patches/llvm-abi-breaking-checks.patch
+++ b/3rd_party/llvm-project/x.x/patches/llvm-abi-breaking-checks.patch
@@ -1,0 +1,41 @@
+diff --git a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+--- a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
++++ b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+@@ -5,2 +5,2 @@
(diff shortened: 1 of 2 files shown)


config/defs.bzl
```diff
@@ -92,6 +92,17 @@ def _declare_sanitizer_config_setting(sanitizer):
     )
 
 def config_settings():
+    bool_flag(
+        name = "assertions",
```
cerisier: I think this may be too broad as it only compiles llvm with assertions not user programs.

dzbarsky: yeah i didn't love the name or this placement. any ideas?

dzbarsky: I guess...should this flag just live inside @llvm-project? :)

cerisier (approved):

PR #615, opened 2026-06-11
Support distributed thinlto on windows

dzbarsky:

PR #617, opened 2026-06-11
[codex] Support distributed ThinLTO on macOS

dzbarsky:
This adds distributed ThinLTO for macOS targets after #615. The existing `windows-thin-lto.patch` remains unchanged, and the macOS rules_cc changes are isolated in `macos-thin-lto.patch`, which is applied after `windows-thin-lto.patch`. The macOS patch corresponds to bazelbuild/rules_cc#748 and only adds Mach-O LLD's direct distributed ThinLTO arguments.

The LLVM source patch adds the exact-file `--lto-obj-path` contract proposed upstream in llvm/llvm-project#203365. The exact-file contract lets the indexing action produce rules_cc's declared `thinlto_merged_object_file` without reproducing Mach-O LLD's `0.<arch>.lto.o` directory naming.

`is_windows_capable_of_thinlto_feature` and `is_macos_capable_of_thinlto_feature` default to false and independently control whether the toolchain exposes the rules_cc `thin_lto` feature on Windows and macOS. `windows_capable_of_thinlto_feature` and `macos_capable_of_thinlto_feature` combine the corresponding build setting with the target OS constraint. `--config=prebuilt` enables both build settings. `is_windows_capable_of_thinlto_feature` can be removed after a rules_cc release contains Windows distributed ThinLTO support. `is_macos_capable_of_thinlto_feature` can be removed after rules_cc and LLVM releases contain macOS distributed ThinLTO support. The previous macOS `-flto=thin` fallback is removed because rules_cc now creates separate indexing and backend actions.

Validation:

- Applied `windows-thin-lto.patch` followed by `macos-thin-lto.patch` to a clean rules_cc checkout and compared the resulting ThinLTO feature with the rules_cc source.
- `bazel test //tests/rule_based_toolchain/legacy_features_as_args:thin_lto_macos_test --test_output=errors` in rules_cc.
- `bazel build --config=prebuilt --config=remote --config=bootstrap //prebuilt/llvm:for_windows_amd64 //prebuilt/llvm:for_windows_arm64 //prebuilt/llvm:for_macos_amd64 //prebuilt/llvm:for_macos_arm64`.
- BuildBuddy invocation: https://app.buildbuddy.io/invocation/ff484673-abb8-41c9-ba4c-a19410241625

### DIFF
--- a/toolchain/BUILD.bazel
+++ b/toolchain/BUILD.bazel
@@ -57,18 +57,35 @@ string_flag(
 )
 
 bool_flag(
-    name = "enable_windows_thin_lto",
+    name = "is_windows_capable_of_thinlto_feature",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+bool_flag(
+    name = "is_macos_capable_of_thinlto_feature",
     build_setting_default = False,
     visibility = ["//visibility:public"],
 )
 
 config_setting(
-    name = "windows_thin_lto_enabled",
+    name = "macos_capable_of_thinlto_feature",
+    constraint_values = [
+        "@platforms//os:macos",
+    ],
+    flag_values = {
+        ":is_macos_capable_of_thinlto_feature": "true",
+    },
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "windows_capable_of_thinlto_feature",
     constraint_values = [
         "@platforms//os:windows",
     ],
     flag_values = {
-        ":enable_windows_thin_lto": "true",
+        ":is_windows_capable_of_thinlto_feature": "true",
     },
     visibility = ["//visibility:public"],
 )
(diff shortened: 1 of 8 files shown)


toolchain/BUILD.bazel
```diff
@@ -62,6 +62,23 @@ bool_flag(
     visibility = ["//visibility:public"],
 )
 
+bool_flag(
```
cerisier, line 65: Will apply both exec and target no :( ?

cerisier (approved):

toolchain/BUILD.bazel
dzbarsky: ah this is named badly. It enables the inclusion of `thin_lto` feature in the toolchain, it doesn't enable the feature by default. Enables user to enable, if you will :) got a better naming idea?

dzbarsky: we could also move these flags to a private package for now, I don't really expect others to use them until rules_cc updates

cerisier: Nah it's ok !

PR #619, opened 2026-06-12
Allow running `AutoconfHdr` action on Windows

nolanderc:
Failed previously due exceeding the maximum argument size for `CreateProcessW`, so we write all arguments to a file instead.


cerisier:
Isn't the whole windows story not supported yet for libstdcxx ?? Or do you need this outside of libstdcxx path ?

nolanderc:
> Isn't the whole windows story not supported yet for libstdcxx ?? Or do you need this outside of libstdcxx path ?

Not sure if you are referring to running with libstdcxx on Windows? I'm only using this for cross-compilation at the moment.

With this fix, I was able to cross-compile libstdcxx for linux from a windows machine. Though I haven't actually done any extensive testing to ensure it actually works as expected. This was just the first blocker I encountered so far.

cerisier:
Oh sorry I thought you were compiling for windows. Because libstdc++ supports windows target but not yet in this project.

dzbarsky:
> > Isn't the whole windows story not supported yet for libstdcxx ?? Or do you need this outside of libstdcxx path ?
> 
> Not sure if you are referring to running with libstdcxx on Windows? I'm only using this for cross-compilation at the moment.
> 
> With this fix, I was able to cross-compile libstdcxx for linux from a windows machine. Though I haven't actually done any extensive testing to ensure it actually works as expected. This was just the first blocker I encountered so far.

This is pretty cool! Want to work through anything else needed and batch it up? Happy to merge together

dzbarsky:
Ok well this doesn't hurt :)

dzbarsky:
Thought I think a proper solution would drop the bash dependency from run_shell :)

PR #625, opened 2026-06-14
Add `module_map_home_cwd` feature

### LINKED ISSUE #624: `-Wincomplete-umbrella` false-positive on exec-config compiles of deeply-nested cc_librarys
I ran into an issue when using bazel-diff. I used Claude to diagnose this, as I understand none of the internas. I would appreciate some help here figuring out wether I am being silly or if it is an actual bug. I don't like disabling checks, so the workaround feels nasty.

## Summary

On macOS (aarch64-darwin), compiling a deeply-nested `cc_library` (e.g. abseil, pulled in by `protoc` when a host tool is built from source) **in the exec configuration** fails with:

```
external/llvm++_repo_rules+llvm-toolchain-minimal-22.1.4-darwin-arm64/module_map.modulemap:4:5:
error: umbrella directory '../../../../../external/llvm++_repo_rules+llvm-toolchain-minimal-22.1.4-darwin-arm64/lib/clang/22' not found
       [-Werror,-Wincomplete-umbrella]
    4 |     umbrella "../../../../../external/llvm++_repo_rules+llvm-toolchain-minimal-22.1.4-darwin-arm64/lib/clang/22"
```

The resource directory **does exist** and is a declared input of the action — the generated module map is fine. The failure is that `clang` resolves the `crosstool` module's relative `umbrella` path from the wrong base directory when the module map is reached through deep, varying-depth transitive `extern module "crosstool"` references. The `-Werror=incomplete-umbrella` added in #394 (on the assumption that an incomplete umbrella means "we generated a bad modulemap") then turns this into a hard error.

The **same** `cc_library` compiles fine in the **target** configuration. Only the exec config (deep host-tool dependency graphs) is affected.

## Environment

- `llvm` module 0.8.3 (pinned `0.7.5`, resolved up); also reproduces on current `main` (`0c1f346`) / v0.8.7 — `runtimes/module_map.bzl` is unchanged since #343 (Feb 2026).
- LLVM 22.1.4, `llvm-toolchain-minimal-22.1.4-darwin-arm64`.
- Host & exec platform: `aarch64-darwin` (macOS, Apple Silicon).
- Bazel 8.7.0. `--spawn_strategy=standalone`.

## Reproduction

Real-world trigger: any host tool built from source that pulls protobuf → abseil, e.g.

```sh
bazel run @bazel-diff//cli:bazel-diff   # bazel-diff (Java) -> protobuf-java -> protoc (exec, C++) -> abseil
```

Minimal repro (forces an exec-config compile of a deeply-nested abseil lib via a tool transition):

```python
# BUILD.bazel
genrule(
    name = "gen",
    outs = ["out.txt"],
    tools = [":host_tool"],          # tools => exec configuration
    cmd = "$(location :host_tool) > $@",
)

cc_binary(
    name = "host_tool",
    srcs = ["host_tool.cc"],
    deps = ["@abseil-cpp//absl/strings:cord"],   # deep transitive cppmap graph
)
```

```sh
bazel build //:gen        # on aarch64-darwin, cold -> -Wincomplete-umbrella error
```

A warm cache or a build that never compiles a host tool from source will not surface it (which is why it's easy to miss locally and only bites cold CI).

## Root cause

`runtimes/module_map.bzl` emits the builtin-headers umbrella as a path **relative to the module map file**:

```python
# _module_map_impl
execroot_prefix = (module_map.dirname.count("/") + 1) * "../"   # module map: bazel-out/<cfg>/bin/external/<repo>/
# _umbrella_submodule
path = execroot_prefix + paths.normalize(directory.path)        # resource dir (source): external/<repo>/lib/clang/22
```

This yields `umbrella "../../../../../external/<repo>/lib/clang/22"`, which is correct **only when clang resolves it relative to the module map's canonical directory** (`bazel-out/<cfg>/bin/external/<repo>/`). We verified the generated map is byte-identical between target and exec configs and resolves correctly from that location.

But `rules_cc` makes every dependent `cc_library`'s generated `.cppmap` declare an `extern module "crosstool"` pointing back at the toolchain module map, with a relative path whose depth depends on where the referencing cppmap sits in the tree. In a deep graph (abseil) we see the **same** module map referenced at multiple depths:

```
extern module "crosstool" "../../../../../../../bazel-out/…/module_map.modulemap"        # 7 × ../
extern module "crosstool" "../../../../../../../../bazel-out/…/module_map.modulemap"      # 8 × ../
extern module "crosstool" "../../../../../../../../../bazel-out/…/module_map.modulemap"   # 9 × ../
```

clang resolves the crosstool module's fixed `../../../../../` umbrella relative to the path through which it loaded the module map. In a deep transitive graph that base is not the canonical `bazel-out/<cfg>/bin/external/<repo>/`, so the umbrella resolves to a directory that does not contain `lib/clang/<v>` → `-Wincomplete-umbrella` → (via #394) `-Werror`.

- **Target config / shallow graphs:** the module map is reached via a short/canonical path → umbrella resolves → OK.
- **Exec config / deep graphs (host tools):** reached via deep, varying-depth references → umbrella mis-resolves → fails.

The resource-dir headers themselves are still found (via the separate `-Xclang -internal-isystem …/lib/clang/<v>/include`); only the module-map umbrella completeness check fails.

## Suggested direction

A relative umbrella path can't be made robust to being loaded through arbitrary-depth `extern module` references, since clang resolves it relative to the referencing path. Possible fixes (you know the toolchain best):

- Reference the builtin resource dir in a way that doesn't depend on the loading context (e.g. avoid an `umbrella` directory for the builtin headers, or express it so it normalizes identically regardless of the `extern module` depth).
- Or reconsider whether `-Werror=incomplete-umbrella` (#394) should be unconditional, given it can fire on correct maps that merely resolve through deep transitive graphs.

## Workaround (for consumers)

Disable `layering_check` for exec/host compiles (target builds keep it):

```
build:macos --host_features=-layering_check
```

Happy to test a patch via `--override_module=llvm=<local checkout>` and provide more traces (clang `-v`, aquery, the generated cppmaps) if useful.

fmeum:
Cleans up the module map generation logic and makes module maps easier to read. May fix #624.

Also remove enabled features from the list of known feature as that's redundant.

### DIFF
--- a/toolchain/cc_toolchain.bzl
+++ b/toolchain/cc_toolchain.bzl
@@ -9,8 +9,6 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
             "@rules_cc//cc/toolchains/args/layering_check:use_module_maps",
             "@llvm//toolchain/features:static_link_cpp_runtimes",
             "@llvm//toolchain/features/runtime_library_search_directories:feature",
-            "@llvm//toolchain/features:archive_param_file",
-            "@llvm//toolchain/features:prefer_pic_for_opt_binaries",
             "@llvm//toolchain/features:parse_headers",
             "@llvm//toolchain/features:external_include_paths",
             # TODO: Restore this after a rules_cc release includes the macOS
@@ -33,17 +31,11 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
         ],
     )
 
+    # Runtime configurations have no opt-in features beyond what is already
+    # enabled, so this set is intentionally empty.
     cc_feature_set(
         name = name + "_runtimes_only_known_features",
-        all_of = [
-            # TODO(zbarsky): Do we want layering check for runtime libs?
-            #"@rules_cc//cc/toolchains/args/layering_check:layering_check",
-            #"@rules_cc//cc/toolchains/args/layering_check:use_module_maps",
-            "@llvm//toolchain/features:archive_param_file",
-            "@llvm//toolchain/features:prefer_pic_for_opt_binaries",
-            # Always last (contains user_compile_flags and user_link_flags who should apply last).
-            "@llvm//toolchain/features/legacy:experimental_replace_legacy_action_config_features",
-        ],
+        all_of = [],
     )
 
     cc_feature_set(
@@ -64,6 +56,7 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
         }) + [
             "@llvm//toolchain/features:prefer_pic_for_opt_binaries",
             "@rules_cc//cc/toolchains/args/layering_check:module_maps",
+            "@llvm//toolchain/features:module_map_home_cwd",
             # These are "enabled" but they only _actually_ get enabled when the underlying compilation mode is set.
             # This lets us properly order them before user_compile_flags and user_link_flags below.
             "@llvm//toolchain/features:opt",
@@ -80,6 +73,8 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
         name = name + "_runtimes_only_enabled_features",
         all_of = [
             "@llvm//toolchain/features:prefer_pic_for_opt_binaries",
+            "@rules_cc//cc/toolchains/args/layering_check:module_maps",
+            "@llvm//toolchain/features:module_map_home_cwd",
             "@llvm//toolchain/features:archive_param_file",
             # Always last (contains user_compile_flags and user_link_flags who should apply last).
             "@llvm//toolchain/features/legacy:experimental_replace_legacy_action_config_features",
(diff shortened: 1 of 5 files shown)


toolchain/cc_toolchain.bzl
```diff
@@ -33,17 +31,11 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
         ],
     )
 
+    # Runtime configurations have no opt-in features beyond what is already
+    # enabled, so this set is intentionally empty.
     cc_feature_set(
```
dzbarsky: should we just remove this target and explicitly pass an empty list at the call site?

dzbarsky (approved):

Yatekii:
I can confirm that this fixes the problem, thanks a lot for the quick resolution!

PR #627, opened 2026-06-14
Resolve tool/module-map helpers at definition site so consumers need no use_repo()

### LINKED ISSUE #621: Composing custom cc toolchains requires use_repo() of the internal toolchain repo (bzlmod "incorrect imports" warning + version-coupled repo name)
When a consumer module composes its own rules_cc rule-based toolchain by reusing
`@llvm`'s public `cc_args` / `tool_map` / module-map helper targets, those helper
labels point into the prebuilt toolchain repository generated by the
`llvm_toolchain_minimal` extension — and they resolve in the *consumer's* repo
namespace. The consumer is therefore forced to import the internal repo directly:

```starlark
llvm_toolchain_minimal = use_extension("@llvm//extensions:llvm_toolchain_minimal.bzl", "llvm_toolchain_minimal")
use_repo(llvm_toolchain_minimal, "llvm-toolchain-minimal-22.1.7-linux-amd64")
```

Since the extension's metadata declares that repo as an indirect dependency,
every fresh Bazel server prints:

```
WARNING: MODULE.bazel:104: The module extension llvm_toolchain_minimal defined in
@llvm//extensions:llvm_toolchain_minimal.bzl reported incorrect imports of
repositories via use_repo():

Imported, but reported as indirect dependencies by the extension:
    llvm-toolchain-minimal-22.1.7-linux-amd64
```

The warning is benign, but the bigger issue is that the imported repo name embeds
the LLVM version and platform (`llvm-toolchain-minimal-<version>-<platform>`), so
every llvm module bump silently breaks downstream MODULE.bazel files until the
pinned name is hand-updated.

### Environment

- Bazel 9.1.1, bzlmod
- llvm module 0.8.6 (LLVM 22.1.7, linux-amd64)

### Suggested fixes

1. Minimal: have `llvm_toolchain_minimal` return `extension_metadata` listing the
   toolchain repo under `root_module_direct_deps`, so direct imports are blessed
   and the warning disappears (the version coupling remains).
2. Preferred: re-export the tool_map / module-map helper targets through the
   `@llvm` module itself (stable aliases under e.g. `@llvm//toolchain/...`), so
   consumers compose toolchains against version-independent labels and never
   need `use_repo()` of the internal repo at all. This removes both the warning
   and the per-bump rename chore.

Happy to test a patch — we carry exactly this pattern (custom toolchain = your
public cc_args + our own sysroots) and can verify against it quickly.

lsjostro:
platform_cc_tool_map / platform_module_map / resource_dir_arg are public: downstream modules compose their own cc_toolchain from them (e.g. with a custom sysroot). They returned bare label strings built from _tool_repo(), so the @llvm-toolchain-minimal-<version>-<platform> repo resolved in the *caller's* repo mapping. That forced every such consumer to use_repo() that repo in their MODULE.bazel — a repo the extension reports as indirect, so bzlmod prints an "incorrect imports" warning, and the version-pinned repo name has to be hand-updated on every LLVM bump.

Wrap the returned labels in Label() so they resolve against this module's own repo mapping at definition site. Consumers no longer need use_repo() and never name the versioned repo. select() keys stay as strings (they are @llvm// / @rules_cc//, already direct deps of any consumer composing a rules_cc toolchain).

Fixes #621.

### DIFF
--- a/toolchain/selects.bzl
+++ b/toolchain/selects.bzl
@@ -2,24 +2,24 @@ LLVM_VERSION = "22.1.7"
 
 def platform_llvm_binary(binary):
     return select({
-        "@llvm//platforms/config:macos_x86_64_prebuilt": "@llvm-toolchain-minimal-%s-darwin-amd64//:bin/%s" % (LLVM_VERSION, binary),
-        "@llvm//platforms/config:macos_aarch64_prebuilt": "@llvm-toolchain-minimal-%s-darwin-arm64//:bin/%s" % (LLVM_VERSION, binary),
-        "@llvm//platforms/config:linux_x86_64_prebuilt": "@llvm-toolchain-minimal-%s-linux-amd64//:bin/%s" % (LLVM_VERSION, binary),
-        "@llvm//platforms/config:linux_aarch64_prebuilt": "@llvm-toolchain-minimal-%s-linux-arm64//:bin/%s" % (LLVM_VERSION, binary),
-        "@llvm//platforms/config:windows_aarch64_prebuilt": "@llvm-toolchain-minimal-%s-windows-arm64//:bin/%s.exe" % (LLVM_VERSION, binary),
-        "@llvm//platforms/config:windows_x86_64_prebuilt": "@llvm-toolchain-minimal-%s-windows-amd64//:bin/%s.exe" % (LLVM_VERSION, binary),
-        "@llvm//toolchain:bootstrapped_toolchain": "@llvm//toolchain/bootstrap:" + binary,
+        "@llvm//platforms/config:macos_x86_64_prebuilt": Label("@llvm-toolchain-minimal-%s-darwin-amd64//:bin/%s" % (LLVM_VERSION, binary)),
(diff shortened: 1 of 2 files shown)


toolchain/selects.bzl
```diff
@@ -28,10 +28,10 @@ def _tool_repo(exec_os, exec_cpu):
     return "@llvm-toolchain-minimal-%s-%s-%s//" % (LLVM_VERSION, os_part, cpu_part)
 
 def platform_module_map(exec_os, exec_cpu):
-    return _tool_repo(exec_os, exec_cpu) + ":module_map"
+    return Label(_tool_repo(exec_os, exec_cpu) + ":module_map")
```
fmeum, line 31: There are more selects further up that could use the same treatment.

lsjostro: sorry! we want to be consistent, now resolved.

fmeum (approved):

dzbarsky (approved):
Thanks!

PR #633, opened 2026-06-18
[codex] Add macOS runners to end-to-end CI

dzbarsky:
By Codex

Add `macos-15` and `macos-15-intel` to `test.matrix.os` in `.github/workflows/ci.yaml`. This runs every end-to-end folder with Bazel 8.x and `last_rc` on both Apple Silicon and Intel macOS runners, extending the existing Linux matrix without changing the Windows or macOS smoke-test jobs.

Validation: parsed `.github/workflows/ci.yaml` with Ruby YAML and ran `git diff --check`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,8 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # only test on linux for now, macos is not ready.
-        os: [ubuntu-latest, ubuntu-24.04-arm]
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -19,8 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # only test on linux for now, macos is not ready.
```
cerisier, line 22: macos 15 ?

dzbarsky: ill do latest

PR #644, opened 2026-06-25
Fix clang-tidy multicall registration

### LINKED ISSUE #640: The `@llvm//tools:clang-tidy` target actually invokes clang
If you invoke something like `@llvm//tools:clang-tidy` you'll see the clang help, not the clang-tidy help. Indeed, the driver doesn't list clang-tidy either and seems to treat the symlink with the same "strip after the dash" behavior as the versioned "clang-22".

```
/external/llvm++llvm_toolchain_minimal+llvm-toolchain-minimal-22.1.7-linux-amd64/bin]$ ./clang-tidy --help | head
OVERVIEW: clang LLVM compiler

USAGE: llvm [options] file...

OPTIONS:
  -###                    Print (but do not run) the commands to run for this compilation
  --analyzer-output <value>
                          Static analyzer report output format (html|plist|plist-multi-file|plist-html|sarif|sarif-html|text).
  --analyze               Run the static analyzer
  -B <prefix>             Search $prefix$file for executables, libraries, and data files. If $prefix is a directory, search $prefix/$file
```

The driver seems to agree that there is no clang-tidy:

```
/external/llvm++llvm_toolchain_minimal+llvm-toolchain-minimal-22.1.7-linux-amd64/bin/llvm --help
OVERVIEW: llvm toolchain driver

USAGE: llvm [subcommand] [options]

SUBCOMMANDS:

  sancov
  symbolizer
  addr2line
  size
  readtapi
  readobj
  readelf
  rc
  windres
  profdata
  objdump
  otool
  objcopy
  bitcode-strip
  install-name-tool
  strip
  nm
  mt
  ml
  lipo
  link
  libtool-darwin
  ifs
  gsymutil
  dwp
  debuginfod-find
  cxxfilt
  c++filt
  cov
  gcov
  cgdata
  ar
  ranlib
  lib
  dlltool
  lld
  ld
  lld-link
  ld.lld
  ld64.lld
  wasm-ld
  dsymutil
  clangd
  clangd
  clang-scan-deps
  clang-format
  clang
  clang++
  clang-cl
  clang-cpp

  Type "llvm <subcommand> --help" to get more help on a specific subcommand

OPTIONS:

  --help - Display this message
```

I'm not sure whether it's actually possible to put clang-tidy into the llvm driver at all. Solutions might be:

- Patch llvm to make it work with the llvm driver. Not sure how much effort that would be.
- Distribute it as a separate binary. Would increase the distribution size. Not sure by how much.
- Alias `@llvm//tools:clang-tidy -> @llvm-project//clang-tools-extra/clang-tidy` and live with custom builds as projects have likely done before.
- If none of the above, remove the symlink and target entirely because it's a (probably fairly low risk because it's linting stage and errors are very obvious) bug.

I really like the idea of having a prebuilt hermetic clang-tidy. Building it from scratch can take a while on local systems, which means people might not want to build/use it, which means AI slops harder and everything explodes.

IIRC clang-tidy significantly benefits from higher optimization levels and things like FDO which already seems to be working for the prebuilds. If this is true, it'd be another big argument for shipping a prebuilt clang-tidy in some form maybe as a separate artifact from the minimal toolchain.

---

(`@llvm` version `0.8.9`)

dzbarsky:
Read selected tool names directly from the driver-tools build setting when generating LLVMDriverTools.def. The label-name reverse map treated //clang-tools-extra/clang-tidy:tool and //clang-tools-extra/clangd:tool as the same target name. Have generate_driver_selects return driver_tool_deps and remove select_driver_tools.

Fixes https://github.com/hermeticbuild/hermetic-llvm/issues/640

### DIFF
--- a/3rd_party/llvm-project/x.x/patches/llvm-driver-tool-order.patch
+++ b/3rd_party/llvm-project/x.x/patches/llvm-driver-tool-order.patch
@@ -1,17 +1,101 @@
+diff --git a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+--- a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
++++ b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+@@ -14 +14 @@
+-load(":driver.bzl", "generate_driver_selects", "generate_driver_tools_def", "llvm_driver_cc_binary", "select_driver_tools")
++load(":driver.bzl", "generate_driver_selects", "generate_driver_tools_def", "llvm_driver_cc_binary")
+@@ -771,3 +771,3 @@
+ # Command line flag to control which tools get included in the llvm driver binary.
+-# The macro also generates config_setting targets used by select_driver_tools().
+-generate_driver_selects(name = "driver-tools")
++# The macro also returns the selected tool dependencies.
++driver_tool_deps = generate_driver_selects(name = "driver-tools")
+@@ -775,5 +775,5 @@
+ generate_driver_tools_def(
+     name = "gen_llvm_driver_tools_def",
+     out = "LLVMDriverTools.def",
+-    driver_tools = select_driver_tools(":driver-tools"),
++    driver_tools = ":driver-tools",
+ )
+@@ -791,5 +791,5 @@
+     deps = [
+         ":Support",
+         ":llvm_driver_tools_def_lib",
+-    ] + select_driver_tools(":driver-tools"),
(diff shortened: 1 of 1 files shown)


cerisier:
Can you make sure this apply cleanly to both 21 and 22 ?

cerisier:
Maybe we should have a quick workflow just to check analysis for prior lllvm versions to validate that patches apply

dzbarsky:
> Maybe we should have a quick workflow just to check analysis for prior lllvm versions to validate that patches apply

yep that's what I was suggesting in the other issue, otherwise I fear we will keep brekaing it

PR #656, opened 2026-06-27
Stop creating symlinked overlay/source view when using LLVM archives

dzbarsky:
The proliferation of llvm-related module extensions (llvm vs llvm_source?) and external repos was pretty confusing (what's an `llvm-raw`??) and the symlink dance made the repo incompatible with R2C2, which means you would still need to materialize the symlink-targeted files out of llvm-project. We can just do 2 targeted bsdtar extractions to create an actual merged layout instead.

from_git and from_local_path continue to work as before - those are mostly for more custom workflows + debugging aids so I am less worried about optimizing them.

The need to vendor the starlark overlay and llvm_configure bits bothers me a bit - open to suggestions how to do it better!


README.md
```diff
@@ -249,33 +249,25 @@ This module allows choosing a specific LLVM version for both the compiler and th
...
+Important: Since this module uses prebuilt compiler archives by default. If you set `llvm.version(...)` to another version, use:
 
 `--@llvm//toolchain:bootstrap_stage=stage1_from_source`
 
 This builds the compiler from source with the stage0 prebuilt seed, which is required when prebuilts for your exact version are not available.
 
 ### Starlark LLVM version variables
 
-If you need LLVM version variables from Starlark, `llvm_source` generates an `llvm_config` repo exposing those values.
-This lets consumers read LLVM version vars without eagerly fetching the entire `@llvm-project` source tree.
```
dzbarsky, line 268: this is no longer a problem since we always fetch the source tree anyway, no more per-project archives

cerisier (approved):
LGTM ! I was concerned at first about the symlink loop but it should only be rare that go further down that top level for `from_path`. 

I wonder if we should consider have a test for from_git and from_path in a later PR since that must become one of the important features I think.

dzbarsky:
> LGTM ! I was concerned at first about the symlink loop but it should only be rare that go further down that top level for `from_path`.
> 
> I wonder if we should consider have a test for from_git and from_path in a later PR since that must become one of the important features I think.

Wdym? I believe from_path behavior is preserved as if was? (But yes we should add a test for this, since we also have broken it in past...)

PR #658, opened 2026-06-27
macOS sanitizers [2/5]: ASan dylib + runtime loading + interface headers

helly25:
**Part 2 of 4** of the macOS compiler-rt sanitizer port — stacked on the foundation PR #657. Resolves #637 (ASan on macOS).

> Diff is cumulative against `main`; the commits new to this PR are the three ASan/toolchain commits at the tip. Review/merge after #657.

On Darwin, ASan ships as `libclang_rt.asan_osx_dynamic.dylib` (a dylib, no static standalone), resolved by the Clang driver from `<resource-dir>/lib/darwin`. This wires it up faithfully (see `lib/asan/CMakeLists.txt` APPLE branch):

- `asan.shared` emits `libclang_rt.asan_osx_dynamic.dylib` with `install_name @rpath/...`, the Darwin link flags (`-stdlib=libc++ -lc++ -lc++abi`) and the `-Wl,-U` weak-symbol flags (asan/lsan/ubsan/sanitizer_common/xray, mirrored in `weak_symbols.bzl`).
- `runtimes:resource_directory` splits into a macOS variant (Darwin `_osx`/`_osx_dynamic` names; only the Darwin-supported runtimes) and the existing default, chosen by an alias; `-resource-dir` is enabled on macOS (link-only).
- The sanitizer interface headers (`sanitizer/*.h`) are wired into the macOS toolchain for user programs (gated to `runtime_stage=complete` to avoid a stage0 sandbox collision) — fixes `'sanitizer/common_interface_defs.h' file not found` from instrumented deps like abseil.
- The asan e2e binary loads the dylib at run time via `dynamic_deps`.

Verified on darwin/arm64: a program including `<sanitizer/common_interface_defs.h>` compiles, links, loads the dylib, and ASan reports `heap-use-after-free`.

PR #291, opened 2026-02-15
[sanitizers] Add support for dfsan aka DataFlowSanitizer

cerisier:

PR #659, opened 2026-06-27
macOS sanitizers [3/5]: UBSan + LSan dylibs

helly25:
**Part 3 of 4** of the macOS compiler-rt sanitizer port — stacked on #658 (ASan).

> Diff is cumulative against `main`; the commits new to this PR are the UBSan/LSan dylibs plus the runtime auto-wiring and per-sanitizer `cc_test`s at the tip. Review/merge after #658.

On Darwin, standalone UBSan and LSan ship only as dylibs (no static standalone runtimes), mirroring ASan (see the APPLE branches of `lib/ubsan/CMakeLists.txt` and `lib/lsan/CMakeLists.txt`):

- `ubsan_standalone.shared` -> `libclang_rt.ubsan_osx_dynamic.dylib` (cxxabi folded in), weak symbols ubsan+sanitizer_common.
- `lsan.shared` -> `libclang_rt.lsan_osx_dynamic.dylib`, weak symbols lsan+sanitizer_common.

Both carry `install_name @rpath/...`, link the SDK libc++, and are placed in the macOS resource directory under their `_osx_dynamic` names.

### Runtime loading on macOS

The dylibs are wired in automatically — targets no longer declare a per-binary `dynamic_deps`:

| Target kind | Link mode | How the runtime dylib is provided |
|---|---|---|
| `cc_test` and other dynamically-linked rules | dynamic | **Toolchain**: the dylib is added to `dynamic_runtime_lib`; the toolchain puts it in the target's runfiles with an `@loader_path` rpath via the `runtime_library_search_directories` feature (now enabled on macOS). Works for any dynamically-linked rule, no per-target wiring. |
| `cc_binary` | static | The `*san_cc_binary` macros add the dylib as a `dynamic_deps` — a statically-linked binary with a dynamically-loaded sanitizer runtime. |

Linux links the static runtimes, so nothing is added there. `--icf=safe` (LLD/ELF only) is also gated out on macOS.

Verified on darwin/arm64: UBSan reports `runtime error: signed integer overflow`, and per-sanitizer `cc_test`s (`asan/ubsan/lsan_runtime_cc_test`) confirm each runtime loads and is functional (ASan poison check, LSan recoverable leak check, UBSan diagnostic path).


3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -1499,6 +1526,59 @@ cc_runtime_stage0_static_library(
...
+        ],
+        "//conditions:default": [],
+    }),
+    shared_lib_name = select({
+        "@platforms//os:macos": "libclang_rt.ubsan_osx_dynamic.dylib",
+        "//conditions:default": "libubsan_standalone.shared.so",
+    }),
+    user_link_flags = select({
+        "@llvm//platforms/config:gnu": [
+            "-L$(location @llvm//runtimes/glibc:glibc_library_search_directory)",
```
dzbarsky: prefer `execpath` to `location` yeah?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -1499,6 +1526,59 @@ cc_runtime_stage0_static_library(
     ],
 )
 
+# On Darwin, standalone UBSan ships as a dylib (libclang_rt.ubsan_osx_dynamic.dylib)
+# with no static standalone runtime; the C++ABI parts (ubsan_cxx) are folded into
+# the dylib. See lib/ubsan/CMakeLists.txt (APPLE branch).
```
dzbarsky: want to link to the file on github?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -2399,6 +2479,58 @@ cc_runtime_stage0_static_library(
...
+            "@llvm//runtimes/glibc:glibc_library_search_directory",
+        ],
+        "@llvm//platforms/config:musl": [
+            "@llvm//runtimes/musl:musl_library_search_directory",
+        ],
+        "//conditions:default": [],
+    }),
+    shared_lib_name = select({
+        "@platforms//os:macos": "libclang_rt.lsan_osx_dynamic.dylib",
+        "//conditions:default": "liblsan.shared.so",
```
dzbarsky, line 2498: ooh did you implement dynamic sanitizers for linux as well along the way? :) if so, can we add some tests?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -2399,6 +2479,58 @@ cc_runtime_stage0_static_library(
...
+    }),
+    user_link_flags = select({
+        "@llvm//platforms/config:gnu": [
+            "-L$(location @llvm//runtimes/glibc:glibc_library_search_directory)",
+            "-lc",
+        ],
+        "@llvm//platforms/config:musl": [
+            "-L$(location @llvm//runtimes/musl:musl_library_search_directory)",
```
dzbarsky: location -> execpath everywhere please

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -3113,6 +3249,24 @@ cc_runtime_stage0_shared_library(
             "-lc",
         ],
         "//conditions:default": [],
+    }) + select({
+        # Darwin: the asan runtime is a dylib loaded via @rpath (the Clang
+        # driver adds -rpath at link time), built against the SDK libc++.
+        # See lib/asan/CMakeLists.txt (APPLE branch) and DARWIN_COMMON_LINK_FLAGS
+        # in cmake/base-config-ix.cmake.
+        "@platforms//os:macos": [
+            "-Wl,-install_name,@rpath/libclang_rt.asan_osx_dynamic.dylib",
```
dzbarsky, line 3259: i wonder if we should make some sort of sanitizier_lib wrapper macro, wdyt?

3rd_party/llvm-project/x.x/compiler-rt/weak_symbols.bzl
```diff
@@ -0,0 +1,125 @@
...
+]
+
+def weak_symbol_link_flags(symbol_lists):
+    """Returns `-Wl,-U,<symbol>` link flags for the given lists of weak symbols.
+
+    Args:
+      symbol_lists: a list of symbol-name lists (e.g. [ASAN_WEAK_SYMBOLS, ...]).
+    """
+    return [
+        "-Wl,-U,{}".format(symbol)
```
dzbarsky: nit: `"-Wl,-U," + symbol` please

e2e/rules_cc/asan_output_test.sh
```diff
@@ -3,11 +3,6 @@ set -euo pipefail
 
 EXPECTED_OUTPUT="ERROR: AddressSanitizer: heap-use-after-free on address"
 
-if [[ "$(uname -s)" == "Darwin" ]]; then
```
dzbarsky: yay!

dzbarsky: but do we actually enable this in CI? Or is that future PR?

dzbarsky: see also https://github.com/hermeticbuild/hermetic-llvm/pull/633

e2e/rules_cc/BUILD.bazel
```diff
@@ -507,6 +507,12 @@ sh_test(
 ubsan_cc_binary(
     name = "ubsan_fail",
     srcs = ["ubsan_fail.cc"],
+    # On macOS the UBSan runtime is a dylib; declare it as a dynamic_dep so it is
+    # in runfiles with an @loader_path rpath and loads at run time.
+    dynamic_deps = select({
```
dzbarsky: this kinda sucks, we already have a flag to enable ubsan, is there any way we can get this wired through `cc_args` somehow so users don't need to do more steps?

e2e/rules_cc/BUILD.bazel
```diff
@@ -705,11 +711,23 @@ exec_test(
...
+    # Declaring it as a dynamic_dep puts it in the binary's runfiles and gives
+    # the binary an @loader_path rpath (via the runtime_library_search_directories
+    # feature), so dyld can load it at run time. On Linux the static runtime is
+    # linked into the binary, so no dynamic_dep is needed.
+    dynamic_deps = select({
+        "@platforms//os:macos": ["@llvm//runtimes/compiler-rt:clang_rt.asan.shared"],
+        "//conditions:default": [],
+    }),
+    linkopts = select({
+        # --icf is an ld.lld (ELF) flag; ld64.lld on macOS does not accept it.
```
dzbarsky, line 756: kinda surprised by this, how did https://github.com/hermeticbuild/hermetic-llvm/blob/68446095b6f16d425b0fe054222343cfbd2f7c1f/toolchain/features/BUILD.bazel#L99 work? Also icf flag is opt-only, are these tests running optimized?

toolchain/args/macos/BUILD.bazel
```diff
@@ -45,6 +45,40 @@ cc_args(
...
+    actions = [
+        "@rules_cc//cc/toolchains/actions:source_compile_actions",
+    ],
+    allowlist_include_directories = [
+        "//sanitizers:sanitizers_headers_include_search_directory",
+    ],
+    # Make sure these sanitizer paths come after any user supplied paths, which
+    # might also include another version of llvm's sanitizer headers.
+    args = [
+        "-Xclang",
```
dzbarsky, line 69: didn't we make these `=`-separted elsewhere? can we check + be consistent?

toolchain/BUILD.bazel
```diff
@@ -154,7 +154,10 @@ cc_args_list(
 cc_args_list(
     name = "resource_dir",
     args = select({
-        "@platforms//os:macos": [],
+        # macOS now ships compiler-rt sanitizer runtimes (e.g. the asan dylib)
```
dzbarsky: pet peeve, the models like to say what happens "now" after a change but it's redundant, can we clean it? :D

e2e/rules_cc/BUILD.bazel
helly25: According to docs: "--icf=safe: Folds only functions where merging them won't break the program's observable behavior. For example, it avoids folding functions if your code explicitly takes their pointer addresses and compares them." So this should be correct, no?

dzbarsky: To clarify I meant we already pass icf flag for Darwin target which seems to contradict the comment you're adding here. Also icf none is default so why did we need to override it anyway?

e2e/rules_cc/BUILD.bazel
dzbarsky: oh I just realized I forgot to submit a draft comment on previous PR...can we add these libs to `runtime_dynamic_lib` filegroup (same as we do for libunwind.so etc), that way rules_cc will automatically add them to link line and also to the rpath? That seems cleaner than each binary needing to link them explicitly! (you'd want to add them to the filegroup conditioned on the `select` of flag being enabled)

e2e/rules_cc/asan_output_test.sh
helly25: The sanitizer tests are enabled in a later PR.

helly25: I guess once this is all in we need to revisit that #633

e2e/rules_cc/BUILD.bazel
helly25: So the earlier setting in `toolchain/features/BUILD.bazel` sets `-icf=safe` (as pointed out by you) but only as a default and not in *san modes. The idea here is to force 'none' mode That works for linux since the last value of the flag wins even across libraries and binaries. And to not have a value for MacOS. No all that said I actually wonder if we need to address the default value at the origin you pointed out and have MacOS always default to empty there?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
helly25: While working on this my Claude ran into an issue when comparing linux/macos. It just kept these since they are supposedly correct. Probably worth tracking after this chain.

e2e/rules_cc/BUILD.bazel
dzbarsky: Right, I'm confused about 2 things:
1) that flag should only be passed with opt, are these tests running with opt?
2) how does it work at all with Darwin target today if the flag isn't recognized?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
dzbarsky: FYI @cerisier

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
helly25: Is that valuable? These (4 so far) are not identical... so I wonder how much value you'd get out of that.

dzbarsky: not sure tbh, it was a bit hard to tell how much overlap or not there was, up to you!

toolchain/args/macos/BUILD.bazel
helly25: Looks the same to me in all places I looked. I wonder buildifier complained a bunch about unsorted lists and you indicate a few things are inconsistent some of which I may introduce. Should we just try a general cleanup up after this chain?

e2e/rules_cc/BUILD.bazel
helly25: I added a new way of exclusion where we check for anything that disallows --icf, either via unsupported mode or os.

e2e/rules_cc/BUILD.bazel
helly25: PTAL, solved as described in PR description.

runtimes/BUILD.bazel
```diff
@@ -64,9 +64,55 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
+config_setting(
```
dzbarsky, line 69: don't think we need this, see below

runtimes/BUILD.bazel
```diff
@@ -64,9 +64,55 @@ filegroup(
...
+    name = "_macos",
+    constraint_values = ["@platforms//os:macos"],
+)
+
+selects.config_setting_group(
```
dzbarsky: replace with something like

```
config_setting(
    name = "_asan_dynamic_runtime_macos",
    constraint_values = ["@platforms//os:macos"],
    flag_values = {"//config:asan": "1"},
)
```

this is much simpler than the macro soup that `selects.config_setting_group` generates :)

toolchain/BUILD.bazel
```diff
@@ -153,15 +153,9 @@ cc_args_list(
...
-        # macOS now ships compiler-rt sanitizer runtimes (e.g. the asan dylib)
-        # in the resource directory, so the Clang driver needs -resource-dir to
-        # find them when linking (e.g. -fsanitize=address). This is a link-only
-        # arg, so it does not affect builtin-header resolution at compile time.
-        "//conditions:default": [
-            "//toolchain/args:resource_dir",
-        ],
-    }),
+    args = [
+        "//toolchain/args:resource_dir",
```
dzbarsky, line 158: heh nice :) actually should this args_list just be inlined into the caller now?

dzbarsky (approved):
thanks for sticking them in the runtime lib, this feels a lot cleaner to me! left a few more nits, lets have @cerisier take a look as well and merge it

3rd_party/llvm-project/x.x/compiler-rt/weak_symbols.bzl
helly25: Done in d14c102 — switched to `"-Wl,-U," + symbol`.

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
helly25: Done in d14c102 — the remaining `-L$(location ...)` (asan.shared's glibc/musl search dirs) now use `$(execpath ...)`. The pre-existing dfsan abilist `$(location)` (unrelated genrule from #291) is left as-is; can fold into a general pass if wanted.

runtimes/BUILD.bazel
helly25: Done in f97015b — replaced the `config_setting_group` trio with one `config_setting` per sanitizer (macOS `constraint_values` + `flag_values`). Used `//config:<san>_target_config: "true"` (the exec-forced wrapper) instead of raw `//config:asan`, so host/exec tools do not pull the dylib when the flag is set globally — the same form `target_<san>_enabled` uses.

runtimes/BUILD.bazel
helly25: Removed — the macOS constraint is folded into each `config_setting` now (f97015b).

toolchain/BUILD.bazel
helly25: `resource_dir` is one of several single-entry `cc_args_list` wrappers here (`stdlib`, `rtlib`, `unwindlib`, `sysroot_flags`, `hermetic_compile_flags`, ...) that re-export a `//toolchain/args:...` target under a local name. Inlining just this one would make it the odd one out; happy to inline them as a set in a post-chain cleanup rather than a one-off.

cerisier (approved):
LGTM. I'll make a pass a different time to move this into a proper runtime stage so that libc and libc++ are not hardcoded in the lib.

Also, I realized that there is a correctness issue with the original CMakeLists because ubsan_standalone is not what was already defined in the compiler-rt.BUILD.bazel ! Will fix it a follow up PR.

dzbarsky:
> LGTM. I'll make a pass a different time to move this into a proper runtime stage so that libc and libc++ are not hardcoded in the lib.
> 
> Also, I realized that there is a correctness issue with the original CMakeLists because ubsan_standalone is not what was already defined in the compiler-rt.BUILD.bazel ! Will fix it a follow up PR.

Well https://github.com/hermeticbuild/hermetic-llvm/pull/666 significantly rewords compiler-rt so make sure you look at that before tweaking more things :)

PR #668, opened 2026-07-02
Fix pic default for musl

keith:
Previously since this defaulted to on, linking cgo with musl would fail
because cgo assumes pic is off by default


keith:
the second commit here shows the issue with this approach. this breaks if you try to enable `--force_pic` unless you also set that for rules_go. this makes it pass:

```
diff --git a/tests/cgo/defs.bzl b/tests/cgo/defs.bzl
index 665010b..811a622 100644
--- a/tests/cgo/defs.bzl
+++ b/tests/cgo/defs.bzl
@@ -7,4 +7,7 @@ force_pic_go_binary, _force_pic_go_binary_internal = with_cfg(
 ).set(
     "force_pic",
     True,
+).set(
+    Label("@rules_go//go/config:linkmode"),
+    "pie",
 ).build()
```

we could maybe make rules_go read the cc* stuff for this, but I'm not sure if it's worth it. the nice thing about this change is that it works by default. we could also continue supporting `//constraints/pie` by making a new default of `unset` or something, and only respecting it when it was unset

keith:
like this works to maintain override behavior:

```patch
diff --git a/constraints/pie/BUILD.bazel b/constraints/pie/BUILD.bazel
new file mode 100644
index 0000000..2542880
--- /dev/null
+++ b/constraints/pie/BUILD.bazel
@@ -0,0 +1,21 @@
+
+constraint_setting(
+    name = "pie",
+    default_constraint_value = "unset",
+)
+
+constraint_value(
+    name = "unset",
+    constraint_setting = "pie",
+)
+
+constraint_value(
+    name = "on",
+    constraint_setting = "pie",
+)
+
+constraint_value(
+    name = "off",
+    constraint_setting = "pie",
+)
diff --git a/toolchain/args/BUILD.bazel b/toolchain/args/BUILD.bazel
index c358fff..f05b12f 100644
--- a/toolchain/args/BUILD.bazel
+++ b/toolchain/args/BUILD.bazel
@@ -153,6 +153,20 @@ cc_args(
     },
 )
 
+cc_args(
+    name = "static_link_executable_custom_pic",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:cpp_link_executable",
+        "@rules_cc//cc/toolchains/actions:lto_index_for_executable",
+        "@rules_cc//cc/toolchains/actions:lto_index_for_dynamic_library",
+        "@rules_cc//cc/toolchains/actions:lto_index_for_nodeps_dynamic_library",
+    ],
+    args = select({
+        "//constraints/pie:on": ["-static-pie"],
+        "//constraints/pie:off": ["-static"],
+    }),
+)
+
 cc_args(
     name = "static_link_executable_pic",
     actions = [
@@ -179,10 +193,13 @@ cc_args(
 
 cc_args_list(
     name = "static_link_executable",
-    args = [
-        ":static_link_executable_pic",
-        ":static_link_executable_no_pic",
-    ],
+    args = select({
+        "//constraints/pie:unset": [
+            ":static_link_executable_pic",
+            ":static_link_executable_no_pic",
+        ],
+        "//conditions:default": [":static_link_executable_custom_pic"],
+    }),
 )
 
 cc_args(
```

keith:
rules_go stuff would be fixed by https://github.com/bazel-contrib/rules_go/pull/4648

toolchain/args/BUILD.bazel
```diff
@@ -154,17 +154,35 @@ cc_args(
...
-        "//constraints/pie:off": ["-static"],
-    }),
+    args = ["-static-pie"],
+    requires_not_none = "@rules_cc//cc/toolchains/variables:force_pic",
+)
+
+cc_args(
+    name = "static_link_executable_no_pic",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:cpp_link_executable",
```
dzbarsky: should we factor out an action_set for these?

constraints/pie/BUILD.bazel
```diff
@@ -1,16 +0,0 @@
-package(default_visibility = ["//visibility:public"])
-
-constraint_setting(
-    name = "pie",
```
dzbarsky, line 4: should we keep these around as no-op with deprecation warning for a bit? give rules_rs and anyone else using them a chance to upgrade?

cerisier (approved):

keith:
i will rebase once rules_go is released with that fix

MODULE.bazel
```diff
@@ -13,6 +13,7 @@ bazel_dep(name = "rules_shell", version = "0.6.1")
 bazel_dep(name = "rules_python", version = "1.7.0")
 bazel_dep(name = "with_cfg.bzl", version = "0.12.0")
 bazel_dep(name = "tar.bzl", version = "0.10.4")
+bazel_dep(name = "rules_go", version = "0.62.0", repo_name = None)
```
keith, line 16: we could just leave this up to the user, but as is this would force the bump if they have a rules_go dep only

fmeum: This is the best way to handle this, no overhead but also "batteries included"

constraints/pie/BUILD.bazel
keith: I put it back but changed the default to unset, in which case we use the new behavior

toolchain/args/BUILD.bazel
keith: I think this is actually wrong since it should never apply to dynamic library links? changing it to all executable links only instead

tests/cgo/defs.bzl
```diff
@@ -0,0 +1,10 @@
+load("@rules_go//go:def.bzl", "go_binary")
+load("@with_cfg.bzl", "with_cfg")
+
+force_pic_go_binary, _force_pic_go_binary_internal = with_cfg(
```
dzbarsky, line 4: I wonder if we should put more of this in some `e2e` module so we don't ship this stuff to users... not a blocker here but I know @cerisier has been wanting to consolidate those/revamp

dzbarsky (approved):

tests/cgo/defs.bzl
keith: `tests` is in `export-ignore`d so this file doesn't get shipped if that's what you mean? IMO the benefit of tests/ is being able to do a single `bazel test ...` without having to do 1 per e2e subdir

dzbarsky: oh cool i forgot we did that. then this is great!

PR #669, opened 2026-07-03
Add patch for letting LLVM 22 compile earlier LLVM versions

sin-ack:
LLVM 22 can no longer build earlier LLVM versions because some things in the LLVM source code are now considered UB and fail at compile-time: https://github.com/llvm/llvm-project/issues/183908

Upstream does not seem to believe in patching earlier versions to fix this so we include this fix here.

Ref: https://github.com/llvm/llvm-project/commit/4bed796a5fea34d38e4754921154b88a99fa6067.patch

### DIFF
--- a/3rd_party/llvm-project/21.x/patches/remove-incorrect-template-specializations.patch
+++ b/3rd_party/llvm-project/21.x/patches/remove-incorrect-template-specializations.patch
@@ -1,0 +1,258 @@
+From 4bed796a5fea34d38e4754921154b88a99fa6067 Mon Sep 17 00:00:00 2001
+From: "A. Jiang" <de34@live.cn>
+Date: Fri, 26 Sep 2025 12:15:48 +0800
+Subject: [PATCH] [CodeGen] Get rid of incorrect `std` template specializations
+
+This patch renames comparators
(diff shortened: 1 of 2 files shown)


3rd_party/llvm-project/21.x/patches/remove-incorrect-template-specializations.patch
```diff
@@ -0,0 +1,258 @@
+From 4bed796a5fea34d38e4754921154b88a99fa6067 Mon Sep 17 00:00:00 2001
```
dzbarsky, line 1: Can you please trim this header down to just an upstream PR link and also remove the tests hunk at the end?

dzbarsky (approved):

sin-ack:
I tried to reproduce this, but even after forcing LLVM 21 and Hexagon in targets, I cannot reproduce the build failure with `@llvm` itself so I'll close this in the interest of not adding unnecessary patches.

ldionne:
> Upstream does not seem to believe in patching earlier versions to fix this so we include this fix here.

It's not that we don't believe in patching earlier versions, it's that the LLVM 21 release had been done for a long time when this issue was brought up in #183908, so patching LLVM 21 was just not an option at this point.

PR #672, opened 2026-07-06
osx: support local macOS SDK

### LINKED ISSUE #67: Support user provided macOS SDK
This is important since we might not always be able to provide a downloadable one.
Also, iOS, VisionOS etc.. are included in the XCode toolchain, not in the downloadable macOS SDK anyway.
Even if we allow auto discovering the SDK from the host, this would only be available on macOS, not Linux for cross-compilation.

### LINKED ISSUE #68: Support host discovered macOS SDK
For the same reasons as #67

enfipy:
Closes #67.
Closes #68.

This PR adds macOS SDK sources to the `osx` module extension (non-hermetic, but the issues are supposed to make this non-hermetic):
1. `osx.from_local(path = ...)` for a user-provided SDK directory
2. `osx.from_host()` for discovering the host SDK with `xcrun --sdk macosx --show-sdk-path`

`osx.from_archive(...)` behavior remains default reproducible path for CI, remote execution, and cross-compilation from non-macOS hosts.

> The new modes are intentionally not marked as reproducible (needed to add extra lines to accommodate that). SDK selected from the host filesystem cannot be made hermetic by this module alone: its contents depend on the installed Xcode/Command Line Tools version, local filesystem state, and host platform. If a team needs hermeticity, the SDK should still be provided as a content-addressed archive via `osx.from_archive(...)`, or through an externally managed reproducible filesystem/artifact provisioning layer.

Motivation: downstream projects (like ZML) are currently blocked when the pinned Apple CLTools SDK package URL becomes unavailable. For example, ZML macOS builds can fail during `@macos_sdk` fetch with Apple CDN `403 Forbidden` / archive timeout errors. With this change, macOS users (and me specifically ;D) who already have Xcode or Command Line Tools installed can opt into `osx.from_host()` or `osx.from_local(...)` and avoid downloading the pinned SDK archive.

Feel free to request or make some changes directly to the branch, as I'm not 100% sure my PR is exactly as you wanted it to resolve.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -242,19 +242,13 @@ jobs:
       - name: bazel test //tests/...
         env:
           USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
-          BUILDBUDDY_API_KEY: ${{ secrets.BUILDBUDDY_API_KEY }}
         shell: bash
         run: |
-          REMOTE_HEADER_ARGS=()
-          if [[ -n "${BUILDBUDDY_API_KEY:-}" ]]; then
-            REMOTE_HEADER_ARGS=(--remote_header="x-buildbuddy-api-key=${BUILDBUDDY_API_KEY}")
-          fi
-
           bazel \
             --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc \
             --bazelrc=.bazelrc \
             test \
-            "${REMOTE_HEADER_ARGS[@]}" \
+            --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7 \
             //tests/...
 
   docs:
--- a/README.md
+++ b/README.md
@@ -232,6 +232,28 @@ Cross-compiling to macOS from any host is supported.
 By default, the official macOS SDK is downloaded from Apple CDN and used hermetically.
 We use a cross-platform reimplementation of `pkgutil` to unpack SDK packages, which works on all hosts.
 
+For local development, the macOS SDK source can be overridden with a local SDK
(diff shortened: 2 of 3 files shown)


dzbarsky:
> For example, ZML macOS builds can fail during @macos_sdk fetch with Apple CDN

Hi :) Just curious, doesn't the [archive.org fallback URL](https://github.com/hermeticbuild/hermetic-llvm/blob/1a8c167ce8b9c4ee36d982e8dcb549a3924e5632/MODULE.bazel#L121-L124) already solve this?

enfipy:
@dzbarsky unfortunately not for me:
- Download from https://swcdn.apple.com/... failed: GET returned 403 Forbidden
- Download from https://web.archive.org/web/... failed: Connect timed out

I will fix CI tomorrow.

enfipy:
I dived deeper about failing `ubuntu-24.04-arm` jobs and they look unrelated to my macOS SDK source changes. They fail with: `FAILED_PRECONDITION: get executor pool info: ARM Linux remote build execution is not enabled for anonymous requests.`

The `tests` job loads `.github/workflows/ci.bazelrc`, which enables remote execution. Unlike the other CI jobs in this workflow, this job was only passing `--remote_header` when `secrets.BUILDBUDDY_API_KEY` was available. On fork PRs that secret is unavailable, so the ARM Linux job attempted BuildBuddy remote execution anonymously and was rejected.

I pushed a small CI-only follow-up that makes this job match the rest of the workflow by passing the existing BuildBuddy key explicitly.

> To be honest, I think it shouldn't expose API key on the workflows in the first place, but I guess it's fine for now :)

enfipy:
Friendly ping @dzbarsky 🙏, if anything else I can fix/rework to help land this PR - I will be more than happy to make this happen.

Issue #675, opened 2026-07-14
bazel coverage silently produces empty reports: no coverage feature, coverage tools undeclared


sombraSoft:
## Summary

The cc_toolchain shipped by this module doesn't support `bazel coverage`. Three independent gaps, each verified on `llvm` 0.8.11 (LLVM 22.1.8, linux-amd64, Bazel 9.1.1):

1. **No instrumentation**: under `bazel coverage`, the toolchain applies zero coverage flags (verified via `bazel aquery 'mnemonic(CppCompile, ...)'` — no `-fprofile*`/`-ftest-coverage`/`-fcoverage-mapping`). Neither the `gcc_coverage_map_format` nor the `llvm_coverage_map_format` feature appears to be implemented, so tests "pass" and the combined lcov report is structurally empty (`SF:` entries, zero `DA:` lines) — a silent failure.
2. **gcov tool broken/undeclared**: the toolchain's declared gcov path (`bin/gcov`, a symlink onto the multiplexed `llvm` binary) is not part of the test runner's inputs, so sandboxed coverage runs fail with `GCov does not exist at the given path: 'external/llvm++llvm_toolchain_minimal+llvm-toolchain-minimal-22.1.8-linux-amd64/bin/gcov'`.
3. **`LLVM_PROFDATA` / `LLVM_COV` not exported**: `rules_cc`'s `collect_cc_coverage.sh` LLVM branch needs the toolchain to provide these; it doesn't, so even manual instrumentation can't be collected without `--test_env` overrides.

Side note: gcov-style instrumentation (`-fprofile-arcs -ftest-coverage`) added manually doesn't link with this toolchain — `ld.lld: error: .eh_frame: relocation is not in any piece` (both `-c opt` and `-c fastbuild`) — so LLVM source-based coverage is the only viable path.

## Reproduction

```sh
# In a module depending on llvm@0.8.11 with @llvm//toolchain:all registered:
bazel coverage --combined_report=lcov -- //some/cc:test
# -> sandboxed: test FAILED, "GCov does not exist at the given path"
# -> unsandboxed: test passes, combined report has 0 DA: lines
```

## Workaround we're using (OpenROAD)

```
coverage:coverage --copt=-fprofile-instr-generate
coverage:coverage --copt=-fcoverage-mapping
coverage:coverage --linkopt=-fprofile-instr-generate
coverage:coverage --experimental_generate_llvm_lcov
[...]

PR #677, opened 2026-07-17
Enable llvm coverage

martin4861:
Enable llvm coverage with following config:
```
coverage --combined_report=lcov
coverage --experimental_generate_llvm_lcov
coverage --experimental_use_llvm_covmap
```

Testing:
- `sh_test` checks for profraw file of a `cc_binary`
- `bazel coverage` integration tests, checks the _coverage_report.dat file

https://github.com/hermeticbuild/hermetic-llvm/issues/675


e2e/rules_cc/defs.bzl
```diff
@@ -237,6 +237,23 @@ profile_cc_binary, _profile_cc_binary_internal = with_cfg(cc_binary).set(
...
+).set(
+    Label("@llvm//config:host_profile"),
+    True,
+).build()
+
+# Builds a cc_binary with llvm source-based coverage instrumentation (-fcoverage-mapping)
+def coverage_cc_binary(name, copts = [], **kwargs):
+    _coverage_cc_binary(
+        name = name,
+        copts = copts + ["-fcoverage-mapping"],
```
keith: shouldn't this come from the toolchain?

e2e/rules_cc/.bazelrc
```diff
@@ -4,3 +4,16 @@ common --config=cc
 common --process_headers_in_dependencies
 common --features=layering_check
 common --modify_execution_info=CppModuleMap=+supports-path-mapping,CppArchive=+supports-path-mapping,CppCompile=+supports-path-mapping
+
+# LLVM source-based coverage (see coverage_lcov_test.sh).
+coverage --experimental_use_llvm_covmap
+coverage --experimental_generate_llvm_lcov
+coverage --combined_report=lcov
+
+# rules_cc's collect_cc_coverage.sh leaves a dangling `gcov` symlink in the
```
keith, line 13: can you link to the fix PR and add the version we can remove this in the comment

martin4861: Added the link, but unsure whether the remark rules_cc>=0.2.23 is the right call?
It seems, that this only works with Bazel 9 and Bazel 8 is failing. Shall we run the tests only on Bazel 9 for now and enable it for Bazel 8, once all PRs are merged?

e2e/rules_cc/defs.bzl
martin4861: Removed. It's already part of the toolchain.

.github/workflows/ci.yaml
```diff
@@ -76,6 +76,14 @@ jobs:
             ${{ matrix.extra_flags }} \
             //...
 
+      - name: bazel coverage e2e (lcov)
+        if: matrix.folder == 'e2e/rules_cc'
+        env:
+          USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
+        working-directory: ${{ matrix.folder }}
+        shell: bash
+        run: ./coverage_lcov_test.sh --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7
```
keith, line 85: token?

e2e/rules_cc/.bazelrc
keith: if it's the same dangling symlink thing I didn't try to backport that to 8.x I don't think, so we'll probably need to keep it to 9.x+

e2e/rules_cc/coverage_output_test.sh
```diff
@@ -0,0 +1,69 @@
...
+TMPDIR="$(mktemp -d)"
+trap 'rm -rf "${TMPDIR}"' EXIT
+
+PROFRAW="${TMPDIR}/coverage.profraw"
+LLVM_PROFILE_FILE="${PROFRAW}" "${BIN}"
+
+if [[ ! -s "${PROFRAW}" ]]; then
+  echo "coverage-instrumented binary did not produce a non-empty .profraw file" >&2
+  exit 1
+fi
```
keith: is this test necessary given the other script that validates things further down the line? if what we want is a test that ensures we pass the coverage command line flags there's probably an easier way to do that

.github/workflows/ci.yaml
martin4861: Usually yes, but this workflow file uses the token in plaintext and does not expose a GH Actions secret.

e2e/rules_cc/.bazelrc
martin4861: It's a different issue: With Bazel 9 the relevant environment variables for the collect_cc_coverage script are properly set in https://github.com/bazelbuild/rules_cc/blob/main/cc/common/cc_helper.bzl#L1052, which are then used in https://github.com/bazelbuild/rules_cc/blob/main/cc/private/coverage/collect_cc_coverage.sh. Afaik, Bazel 8 uses the native cc rules and therefore does not set the env vars.

I will restrict coverage to Bazel 9 for now.

e2e/rules_cc/coverage_output_test.sh
martin4861: The end to end test which runs `bazel coverage` is sufficient. Imho we can remove this test.

.github/workflows/ci.yaml
keith: oh my bad ha

keith (commented):
lgtm once you toss that test

e2e/rules_cc/.bazelrc
keith: ah right that change, for sure will be 9.x+ only then

.github/workflows/ci.yaml
```diff
@@ -176,6 +184,19 @@ jobs:
...
+        working-directory: ${{ matrix.folder }}
+        shell: bash
+        run: |
+          bazel \
+            --bazelrc=.bazelrc \
+            test \
+            --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7 \
+            //:coverage_output_test
+          ./coverage_lcov_test.sh --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7
+
```
keith:
```suggestion
```

keith (approved):

PR #681, opened 2026-07-20
Support producing interface libraries

### LINKED ISSUE #219: investigate llvm-ifs and cc build_interface_libraries feature

keith:
Interface libraries are shared library stubs only used during linking.
This can massively reduce the bytes transferred for shared library links
since these stubs only have to contain enough for the public API, not
everything else. Today llvm-ifs does not correctly produce stubs for
defined versioned symbols in a library (which is likely very rare for
bazel). In that case we just copy the original library instead.

Fixes https://github.com/hermeticbuild/hermetic-llvm/issues/219

### DIFF
--- a/tests/interface_libraries/BUILD.bazel
+++ b/tests/interface_libraries/BUILD.bazel
@@ -1,0 +1,24 @@
+load(":interface_library_test.bzl", "interface_library_test", "test_shared_library")
+
+test_shared_library(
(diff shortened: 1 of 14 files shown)


dzbarsky (approved):
Did it work for Chris?

fmeum (approved):

keith:
Chris can't test yet

PR #683, opened 2026-07-23
Add empty targets for mismatched kernel headers

keith:
Without this you can't do queries on mismatched platforms

For example:

```
bazel query 'rdeps(//..., @llvm//...)'
```


dzbarsky:
Closing in favor of https://github.com/hermeticbuild/hermetic-llvm/pull/684

PR #684, opened 2026-07-23
Fix kernel header queries for unavailable architectures

dzbarsky:
## Summary

Fix `bazel query` failures for `@linux_kernel_headers_riscv.4.14.336` and other unavailable kernel-header archives. `bazel query` traverses every `select()` branch, including RISC-V and ARM64 repositories that the kernel-header index never creates for older Linux versions.

- Generate `LINUX_KERNEL_ARCHS_BY_VERSION` from the kernel-header archive index.
- Generate explicit-kernel and libc selectors only for existing architecture/version combinations.
- Test the RISC-V 4.14/4.15 and ARM64 3.6/3.7 boundaries.
- Run the original failing dependency queries in the existing Bazel 8 and release-candidate CI matrix.

## Validation

- Original query passes on Bazel 8 and Bazel 9.
- All 8 `//tests/...` tests pass under Linux remote execution.
- Bazel 8 kernel-header regression test passes.
- Changed Bazel and Starlark files pass buildifier.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -254,7 +254,7 @@ jobs:
           restore-keys: |
             bazel-cache-tests-${{ matrix.os }}-${{ matrix.bazel_version }}
 
-      - name: bazel test //tests/...
+      - name: bazel query kernel headers and test //tests/...
         env:
           USE_BAZEL_VERSION: ${{ matrix.bazel_version }}
           BUILDBUDDY_API_KEY: ${{ secrets.BUILDBUDDY_API_KEY }}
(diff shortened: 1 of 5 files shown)


cerisier:
Are those legitimately missing or should something be done in Linux headers repo ?

dzbarsky:
> Are those legitimately missing or should something be done in Linux headers repo ?

I think it depends on the kernel version, but I forget how we tie that to glibc version, I'm sure you remember. Here's what codex said:

image

so i think this is fine for now but it would be more correct to enable them for the newer kernel header repos, yeah?

cerisier:
Yup ! The match is outdated now that we support per distro constraints. I'll simplify it with latest kernel headers asap. Good to merge

PR #687, opened 2026-07-27
Fix non-hermetic golang strip targeting macOS

keith:
Previously this fell back to using /bin/strip on linux, which likely
doesn't support stripping Mach-O binaries. It shouldn't be using a
system binary for this anyways.

The `/bin/strip` path comes from golang calling
`<linker driver> --print-prog-name strip`, so now we return the correct
value for this

### DIFF
--- a/tests/cgo/BUILD.bazel
+++ b/tests/cgo/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@bazel_lib//lib:transitions.bzl", "platform_transition_binary")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_go//go:def.bzl", "go_binary")
-load(":defs.bzl", "force_pic_go_binary")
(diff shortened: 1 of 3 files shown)


dzbarsky:
Wow, nice find! I am wondering if we can pass `-B./execroot/path/to/bin` or something like that, did you happen to give it a shot? I'm having codex play around and see if that can help

cerisier (approved):
The level of depth this project gets us lol

dzbarsky:
woah look at what we found, @keith @cerisier check this out - https://github.com/llvm/llvm-project/blob/5e91f5d57a19752fe245ab64c1265e26c44d0d76/clang/lib/Driver/Driver.cpp#L1626-L1627

WDYT of doing https://github.com/hermeticbuild/hermetic-llvm/pull/689 instead?

dzbarsky:
We took the other one :)

PR #689, opened 2026-07-28
Fix hermetic macOS cgo stripping with COMPILER_PATH

dzbarsky:
Alternative to #687 using the same macOS cgo regression test.

Go discovers Darwin post-link tools by invoking the external linker with `--print-prog-name strip`, without passing external linker flags. Set `COMPILER_PATH` to the hermetic LLVM `llvm-` prefix for prebuilt and bootstrapped toolchains so Clang resolves `llvm-strip` itself. This leaves `tools/internal/link_wrapper.c` unchanged.

`COMPILER_PATH` is not GCC-only: [LLVM 22.1.8 Clang adds both `-B` and `COMPILER_PATH` to the same `PrefixDirs`](https://github.com/llvm/llvm-project/blob/llvmorg-22.1.8/clang/lib/Driver/Driver.cpp#L1612-L1623). Resolve the prebuilt prefix from the `bin` directory so Windows `llvm.exe` does not produce an invalid `llvm.exe-` prefix. Add the existing `llvm` multicall executable to the bootstrap tool declarations for the same lookup in bootstrapped toolchains.

Validation:

- `bazel test --config=remote --remote_download_outputs=minimal --jobs=128 --lockfile_mode=off //tests/cgo/...`: all six tests pass, including `hello_cgo_macos_arm64_test`.
- [Remote test run](https://app.buildbuddy.io/invocation/1ab2b2f0-0781-4ddc-9d9e-b3a9a33ea326).
- Buildifier and `git diff --check` pass.

### DIFF
--- a/tests/cgo/BUILD.bazel
+++ b/tests/cgo/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@bazel_lib//lib:transitions.bzl", "platform_transition_binary")
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@rules_go//go:def.bzl", "go_binary")
-load(":defs.bzl", "force_pic_go_binary")
+load(":defs.bzl", "force_pic_go_binary", "macos_arm64_debug_go_binary")
 
 go_binary(
     name = "hello_cgo",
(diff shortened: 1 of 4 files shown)


keith (approved):
that's an old one! https://github.com/llvm/llvm-project/commit/823011606603b43f21a8be900d0ca650b675a9aa

cerisier (approved):

PR #690, opened 2026-07-29
Use a real directory for hermetic Clang strip discovery

dzbarsky:
Follow up to #689.

Go's macOS external linker discovers `strip` by running Clang with `--print-prog-name strip`. Clang interprets an existing `COMPILER_PATH` directory as a directory containing an executable named `strip`, but the prebuilt and bootstrap toolchains provide `llvm-strip`. The existing `{llvm}-` and `{llvm_bin}/llvm-` settings therefore rely on Clang's fallback behavior for a nonexistent executable prefix and concatenate text onto a Bazel-formatted executable path.

Use the existing `cc_tool` to create a lightweight executable alias named `strip` (or `strip.exe` on Windows), use Skylib `directory` and `subdirectory` to expose its actual directory as `DirectoryInfo`, and set `COMPILER_PATH` to `{llvm_bin}` for both prebuilt and bootstrap toolchains. Add each directory to the link wrapper's action inputs and remove the bootstrap `llvm` executable introduced only for the previous `{llvm}-` setting. Keep the existing macOS cgo regression test and `tools/internal/link_wrapper.c` unchanged.

Validation:

- [The unchanged macOS cgo regression fails when `COMPILER_PATH` is removed](https://app.buildbuddy.io/invocation/4646d6af-5da2-4102-89aa-bd206c1ce9b7): `exec: "strip": executable file not found in $PATH`.
- [All six existing cgo tests pass with the real-directory fix](https://app.buildbuddy.io/invocation/c5bd86e8-7628-4eed-91db-7382926313f6).
- Buildifier and `git diff --check` pass.


cerisier (commented):
Is strip the only binary concerned by compiler path ? Will this affect discovery of other orchestrated binaries ?

PR #700, opened 2026-08-11
Support libFuzzer on macOS

martin4861:
Add the libFuzzer runtimes (`libclang_rt.fuzzer_osx.a`, `fuzzer_no_main_osx`, `fuzzer_interceptors_osx`) to the Darwin resource directory, so `-fsanitize=fuzzer` links on macOS instead of failing with a missing archive and undefined main.
Enables the three existing e2e/rules_cc fuzzer tests on macOS.


dzbarsky (approved):

cerisier (approved):
Cool ! I forgot that this was already ported to macos !! Would be cool to find a solution next to not require dynamic deps. maybe the new cc runtime api ?

Issue #434, opened 2026-04-10
`llvm-libc` support


mattyclarkson:
A system root that uses [`llvm-libc`](https://github.com/llvm/llvm-project/tree/main/libc). [Relevant talk](https://www.youtube.com/watch?v=HtCMCL13Grg).

cerisier:
We definitely want to support this. Also because of the recent GPU support we're working on. I can probably find some cycles to work on this in the coming weeks, but if anyone (you) want to give it a go, i'm happy to provide support and guidance for navigating the practices of this project.

PR #704, opened 2026-08-14
Support freestanding AArch64

theoparis:
This adds support for the `aarch64-unknown-none-elf` target triple.

My experimental Fuchsia fork's  zircon kernel expects the the libcxx headers to be provided on top of the ktl headers, but it only ever uses the freestanding headers.

https://github.com/hermeticbuild/hermetic-llvm/issues/434 would also be helpful here.

### DIFF
--- a/3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
+++ b/3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
@@ -651,6 +651,11 @@ cc_library(
             "COMPILER_RT_HAS_FLOAT16",
         ],
         "//conditions:default": [],
+    }) + selects.with_or({
+        (
+            "@llvm//platforms/config:none_aarch64",
+        ): ["ENABLE_BAREMETAL_AARCH64_FMV"],
+        "//conditions:default": [],
     }),
     textual_hdrs = [
         "lib/builtins/fp_add_impl.inc",
@@ -699,6 +704,9 @@ cc_library(
         ],
         "@llvm//platforms/config:windows_aarch64": [
             "lib/builtins/cpu_model/aarch64/fmv/windows.inc",
+        ],
+        "@llvm//platforms/config:none_aarch64": [
+            "lib/builtins/cpu_model/aarch64/fmv/baremetal.inc",
         ],
         "//conditions:default": [],
     }),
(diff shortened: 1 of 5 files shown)


3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -700,6 +706,9 @@ cc_library(
         "@llvm//platforms/config:windows_aarch64": [
             "lib/builtins/cpu_model/aarch64/fmv/windows.inc",
         ],
+        "@platforms//cpu:aarch64": [
```
cerisier: most likely need the config none_aarch64 !

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -651,6 +651,12 @@ cc_library(
             "COMPILER_RT_HAS_FLOAT16",
         ],
         "//conditions:default": [],
+    }) + selects.with_or({
+        (
+            "@platforms//os:none",
+            "@platforms//cpu:aarch64",
```
cerisier: same here, instead of select.with_or.

toolchain/BUILD.bazel
```diff
@@ -323,6 +323,14 @@ cc_args_list(
     ],
 )
 
+# Callers are expected to provide their own runtime support, e.g. `operator new`) so freestanding headers work
+cc_args_list(
+    name = "none_aarch64_toolchain_args",
```
cerisier: Will probably need proper generic treatment vs aarch64 only.

cerisier (commented):
Thanks for the PR.

Am I right to understand that you are compiling for a hosted, non traditional OS? i.e not freestanding ?

If so, we might be able to enrich this PR for the general hosted "none" model. 

If you are trying freestanding, then I believe this will need quite more work, especially wrt to compiler invocation but also non hosted c++ headers and library targets.

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -651,6 +651,11 @@ cc_library(
             "COMPILER_RT_HAS_FLOAT16",
         ],
         "//conditions:default": [],
+    }) + selects.with_or({
```
dzbarsky, line 654: don't think we need `selects.wiht_or` here as you only have a single branch, just do a normal `select`!

PR #718, opened 2026-08-26
Add Windows MSVC target support

cerisier:
## Summary

This PR adds hermetic LLVM target support for the Windows MSVC ABI on x86-64 and ARM64.

It allows Bazel C/C++ actions running on Linux execution platforms to use LLVM's `clang-cl`, `llvm-ar`, and `lld-link` tools to produce native Windows PE/COFF programs. The resulting executables, libraries, DLLs, and import libraries use the Microsoft ABI, Microsoft CRT, and Windows SDK while continuing to use static libc++ as the C++ standard library.

This is intentionally the **target-support** boundary. It makes Windows MSVC a supported compilation target; it does not yet publish a Windows LLVM prebuilt or make Windows a supported execution host for this LLVM toolchain.

## Contract and scope

- Target OS, target ABI/runtime, compiler personality, and execution platform remain separate axes.
- Windows MSVC targets use the clang-cl driver personality and LLVM tools only: `clang-cl`, `llvm-ar`, and `lld-link`.
- The Microsoft VC runtime and Windows SDK are declared Bazel inputs. They are fetched only after the Microsoft EULA repository flags are explicitly accepted; ambient Visual Studio and SDK installations are not used by compilation actions.
- Static libc++ remains the C++ standard library. Microsoft STL is not introduced.
- Ordinary MSVC consumers default to `/MD`; `/MT` is available through the existing CRT features. Release products can select `/MT` without changing the consumer default.
- Existing MinGW, Linux, and macOS toolchains retain their existing driver grammar, runtime selection, and behavior.
- Target actions remain suitable for hermetic Linux remote execution.

### Windows MSVC target platforms and toolchain selection

- Add explicit x86-64 and native ARM64 Windows MSVC target platforms and constraints.
- Register compatible target toolchains for Linux x86-64 and Linux ARM64 execution platforms.
- Keep MinGW and MSVC as distinct Windows ABI/runtime choices rather than treating clang-cl as an operating-system abstraction.

### clang-cl and COFF action semantics

- Add clang-cl compile, assemble, archive, link, DLL, import-library, PDB, and ThinLTO action protocols.
- Model MSVC/COFF machine selection, object and library suffixes, subsystem selection, deterministic object/link behavior, response-file handling, and default system libraries in their semantic argument groups.
- Model `/MD`, `/MT`, and debug CRT selection with validation that rejects contradictory or incomplete CRT configurations.
- Keep compiler-personality behavior in the clang-cl layer and Windows ABI/CRT/SDK behavior in the Windows MSVC layer.

### Hermetic SDK and runtime closure

- Declare Microsoft VC runtime and Windows SDK repositories, inputs, include paths, library paths, and system libraries.
- Build compiler-rt builtins and static libc++ for both MSVC target architectures.
- Carry the supported LLVM overlay adjustment needed to avoid the duplicate `std::nothrow` definition when static libc++ is linked with the Microsoft runtime.
- Preserve compiler-rt builtins as transitive toolchain inputs.

### Portable construction tools

- Add portable C implementations of the case-insensitive copy and VFS helpers used to construct Windows-compatible SDK/header views on case-sensitive execution hosts.
- Add a portable COFF DEF parser, based on Bazel's supported behavior, for generated import libraries.
- Test those helpers independently and make them available without requiring an already runtime-complete C++ bootstrap toolchain.

### End-to-end target proof

- Add analysis and action-graph checks for target triples, clang-cl grammar, declared inputs/tools, CRT modes, deterministic archives, DLLs/import libraries, PDBs, generated DEF files, and ThinLTO.
- Add artifact checks for AMD64 and native ARM64 PE/COFF outputs, including the negative ARM64EC boundary.
- Build representative programs on Linux RBE and execute `/MT` outputs on the corresponding native Windows x86-64 and ARM64 CI runners.
- Exercise Hello World output, generated DEF consumption, DLL/import-library behavior, libc++ exceptions/RTTI/allocation/streams/filesystem/threading, C and assembly inputs, alwayslink archives, and ThinLTO.

## Not included

- No Windows LLVM prebuilt archive is produced, registered, uploaded, or published by this PR.
- No native Windows execution toolchain is registered. Compiler and linker actions still execute with the existing Linux-hosted LLVM tools; Windows CI only executes the resulting target programs.
- No Windows-hosted LLVM bootstrap or release build is introduced here.
- MSVC FDO profile application remains unsupported at this target-only boundary. ThinLTO is supported.
- Microsoft redistributable DLLs are not packaged. `/MD` action and import behavior are covered, while runtime execution uses `/MT` so CI does not depend on redistributing those DLLs.
- Dynamic libc++ and Microsoft STL are outside the supported contract; the toolchain uses static libc++.
- Upstreaming the reusable clang-cl/rules_cc pieces is separate from establishing and releasing the complete downstream behavior.

## Tests

- Windows MSVC action and analysis suites for both target architectures, including negative CRT, unsupported-feature, source-bootstrap, and FDO boundaries.
- `//:windows_msvc_artifacts_matrix` for x86-64 and ARM64, `/MD` and `/MT`, archives, DLLs/import libraries, generated DEF files, PDBs, ThinLTO, and PE/COFF inspection.
- `//tools/case_insensitive_copy:all`, `//tools/case_insensitive_vfs:all`, and `//tools/coff_def_parser:all`.
- Multi-target remote cross-compilation from macOS through Linux RBE.
- Linux-RBE-built `/MT` x86-64 programs executed under Wine: Hello World, generated DEF consumption, DLL behavior, threaded libc++ behavior, and ThinLTO behavior. BuildBuddy invocation: `f9313855-16fa-4ad6-a090-099db8fafdb0`.
- The existing Windows x86-64 and ARM64 CI jobs build the same MSVC target programs through Linux RBE and execute them natively.
- Gazelle diff, public Starlark documentation generation, and buildifier.

## Next

1. Merge this target-support PR as the semantic and action-protocol foundation.
2. Add the Windows prebuilt producer and release changes in a dedicated PR. The Windows release archive should contain the pieces needed to support both the existing MinGW toolchain and the new MSVC toolchain as distinct registrations.
3. Publish that prebuilt release only after both architectures and both Windows target personalities pass archive inspection and light runtime checks.
4. Add the native Windows execution/host toolchain in a separate PR, consume the published prebuilt, and add a Windows compilation matrix row that compiles and runs programs with the MSVC toolchain directly on Windows.
5. Prepare focused upstream LLVM or rules_cc changes only after the released downstream path and Windows-hosted toolchain behavior are proved end to end.


toolchain/runtimes/cc_runtime_copy_file.bzl
cerisier, line 1: Why do we still need this ? aren't all the transitions already applied in the target that we would copy ?

toolchain/cc_toolchain.bzl
```diff
@@ -129,27 +192,158 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
...
+            "@llvm//toolchain:runtimes_stage1": ["@llvm//toolchain/runtimes:toolchain_args"],
+            "@llvm//toolchain:runtimes_stage1_hosted": ["@llvm//toolchain/runtimes:toolchain_args"],
+            "//conditions:default": ["@llvm//toolchain:toolchain_args"],
+        }) + select({
+            # ABI validation is a toolchain invariant, not a user-disableable
+            # feature.
+            "@llvm//constraints/windows/abi:msvc": ["@llvm//toolchain/features/msvc:configuration_validation_args"],
+            "//conditions:default": [],
+        }) + select({
+            "@llvm//constraints/windows/abi:msvc": ["@llvm//toolchain/features/msvc:crt_compile_args"],
```
cerisier: No way to get that inside the toolchain:BUILD.bazel windows toolchain args ?

toolchain/cc_toolchain.bzl
```diff
@@ -129,27 +192,158 @@ def cc_toolchain(name, tool_map, module_map = None, extra_args = []):
...
+            "//conditions:default": [
+                # TODO: rules_cc passes extra args to these actions, ideally these would be fixed in rules_cc.
+                "@llvm//toolchain/args:ignore_unused_command_line_argument",
```
cerisier: Same comment, but also not idea to have this globally unless if was like that before. Any solution ?

3rd_party/llvm-project/21.x/patches/libcxx-vcruntime-nothrow.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/libcxx/src/new_helpers.cpp b/libcxx/src/new_helpers.cpp
+--- a/libcxx/src/new_helpers.cpp
++++ b/libcxx/src/new_helpers.cpp
+@@ -12,7 +12,7 @@
+ namespace std { // purposefully not versioned
+-
++
+-#ifndef __GLIBCXX__
++#if !defined(__GLIBCXX__) && !defined(_LIBCPP_ABI_VCRUNTIME)
```
dzbarsky: we upstreaming these? :)

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -26,6 +26,42 @@ load("@llvm//toolchain/runtimes:cc_unsanitized_library.bzl", "cc_unsanitized_lib
 load("@llvm-project//:vars.bzl", "LLVM_VERSION_MAJOR")
 load("@rules_cc//cc:cc_library.bzl", "cc_library")
 
+selects.config_setting_group(
```
dzbarsky: I wonder if all these should go somewhere under `//private/constraints` instead, do we reuse them across various BUILD files?

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
```diff
@@ -26,6 +26,42 @@ load("@llvm//toolchain/runtimes:cc_unsanitized_library.bzl", "cc_unsanitized_lib
...
+    name = "windows_aarch64_msvc_optimized",
+    constraint_values = [
+        "@llvm//constraints/windows/abi:msvc",
+        "@platforms//cpu:aarch64",
+        "@platforms//os:windows",
+    ],
+    flag_values = {"@llvm//config:runtimes_optimization_mode": "optimized"},
+)
+
+selects.config_setting_group(
```
dzbarsky: nit: seems a little silly to define this `or` when you can just do 

```
config_setting(
    name = "windows_msvc_optimized",
    constraint_values = [
        "@llvm//constraints/windows/abi:msvc"
        "@platforms//os:windows",
    ],
    flag_values = {"@llvm//config:runtimes_optimization_mode": "optimized"},
)
```

- this will generate fewer targets :) if you don't like the verbosity, we can make a small macro:
```
def windows_msvc_config_setting(name, extra_constraints):
config_setting(
    name = name,
    constraint_values = [
        "@llvm//constraints/windows/abi:msvc"
        "@platforms//os:windows",
    ] + extra_constraints,
    flag_values = {"@llvm//config:runtimes_optimization_mode": "optimized"},
)
```

same with the config_setting_group above IMO!

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -59,22 +59,63 @@ _CONFIG_SITE_THREAD_API_PTHREAD_LINES = [
...
+]
+
+_CONFIG_SITE_ABI_DEFAULT_LINES = [
+    "#define _LIBCPP_ABI_FORCE_ITANIUM       0",
+    "#define _LIBCPP_ABI_FORCE_MICROSOFT     0",
+    "#define _LIBCPP_HAS_TIME_ZONE_DATABASE  1",
+]
+
+_CONFIG_SITE_LLVM_LINE_LINES = ([
+    "// LLVM 23 libcxx/include/__config_site.in contract.",
```
dzbarsky, line 92: ugh "contract" is a slop word for me now :)

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -292,36 +363,76 @@ cc_library(
...
+            "/clang:-Xclang",
+            "/clang:--dependent-lib=msvcprt",
+        ],
+        "//conditions:default": [],
+    }),
+    features = select({
+        # clang-cl discovers this archive through its configured library
+        # search path, outside Bazel's LTO compilation context. Keep its
+        # members native so ThinLTO imports remain declared action inputs.
+        ":windows_msvc": ["-thin_lto"],
```
dzbarsky, line 429: hmm I didn't follow this reasoning "members native"? Is the TL;DR that we are not gonna LTO libcxx on msvc? maybe we can reword this to be clearer at least?

e2e/rules_cc/windows_msvc_action_test.sh
```diff
@@ -0,0 +1,314 @@
...
+assert_absent "${action_dir}/dbg-compile-flags.txt" "/D_DEBUG"
+assert_contains "${action_dir}/archive.txt" "llvm-ar"
+assert_contains "${action_dir}/archive.txt" "rcsD"
+assert_contains "${action_dir}/archive.txt" "windows_msvc_libcxx_behavior_support.lib"
+assert_absent "${action_dir}/archive.txt" "llvm-lib"
+assert_absent "${action_dir}/staged-runtime-validation.txt" "Mnemonic: ValidateStaticLibrary"
+assert_contains "${action_dir}/complete-runtime-validation.txt" "Mnemonic: ValidateStaticLibrary"
+assert_contains "${action_dir}/complete-runtime-validation.txt" "static-library-validator"
+
+assert_contains "${action_dir}/link.txt" "bin/clang-cl"
```
dzbarsky, line 332: ha! nice guardrails. Do you think these could be analysis_test instead? Or we need to do it like this?

runtimes/msvc/BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+
+validated_directory_info(
+    name = "msvc_include",
+    src = "@msvc_runtime//:msvc_include",
+    expected_path_suffix = "/Contents/VC/Tools/MSVC/{}/include".format(MSVC_PAYLOAD_VERSION),
+)
+
+[
+    validated_directory_info(
+        name = "winsdk_{}_include".format(directory),
```
dzbarsky: nit: %s over format - faster and shorter ;)

runtimes/msvc/versions.bzl
```diff
@@ -0,0 +1,6 @@
+"""Pinned Microsoft payload and clang-cl compatibility versions."""
+
+MSVC_PAYLOAD_VERSION = "14.50.35717"
```
dzbarsky: should this be generated by @windows_support?

tests/linux_linking/BUILD.bazel
```diff
@@ -80,11 +81,14 @@ sh_test(
     ],
     data = [
         ":hello_x86_64_musl_pie_on",
+    ],
+    rule = sh_test,
```
dzbarsky: usually we do 
```
exec_test(
    sh_test,
    name = ....
)
```

makes it easier to see wtf the test is IMO :)

toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -0,0 +1,595 @@
...
+        "/clang:-fthin-link-bitcode={lto_indexing_bitcode_file}",
+    ],
+    format = {
+        "lto_indexing_bitcode_file": "@rules_cc//cc/toolchains/variables:lto_indexing_bitcode_file",
+    },
+    requires_not_none = "@rules_cc//cc/toolchains/variables:lto_indexing_bitcode_file",
+)
+
+cc_args(
+    name = "thin_lto_indexing_param_args",
```
dzbarsky, line 122: these bits we can upstream to rules_cc right?

toolchain/features/legacy/BUILD.bazel
```diff
@@ -133,6 +250,59 @@ cc_feature_set(
...
+
+cc_feature_set(
+    name = "clang_cl_windows_msvc_legacy_replacements",
+    all_of = [
+        # Target ABI/tool protocol: llvm-ar writes deterministic COFF archives.
+        "//toolchain/features/msvc:archiver_flags",
+        # Target-inapplicable under COFF: registered empty implementations.
+        "//toolchain/features/msvc:pic_flags",
+        # Driver-dialect replacements, kept in the same order as the generic
+        # rules_cc spine above.
```
dzbarsky, line 281: omg what is a "spine", these clankers...

toolchain/declare_toolchains.bzl
```diff
@@ -44,3 +56,23 @@ def declare_toolchains(*, execs = SUPPORTED_EXECS, targets = SUPPORTED_TARGETS):
                 toolchain_type = "@bazel_tools//tools/cpp:toolchain_type",
                 visibility = ["//visibility:public"],
             )
+
+            if target_os == "windows" and (exec_os, exec_cpu) in MSVC_TARGET_SUPPORTED_EXECS:
+                native.toolchain(
+                    name = "{}_{}_to_{}_{}_msvc".format(exec_os, exec_cpu, target_os, target_cpu),
+                    exec_compatible_with = [
+                        "@platforms//cpu:{}".format(exec_cpu),
```
dzbarsky: nit: `@platforms//cpu:" + exec_cpu` please :) here and everywhere

MODULE.bazel
```diff
@@ -105,6 +106,42 @@ use_repo(musl, "musl_libc")
...
+    visual_studio_installer_manifest_url = "https://download.visualstudio.microsoft.com/download/pr/fdc37f6e-59f6-4054-838a-b476eeaa6ec3/1d82370739911457e0a2f6be15d8b5f569531b352b2eabb961429eea1e99e356/VisualStudio.vsman",
+)
+use_repo(msvc_runtime, "msvc_runtime")
+
+windows_sdk = use_extension(
+    "@windows_support//windows:extensions.bzl",
+    "windows_sdk",
+)
+windows_sdk.configure(
+    transformations = {
```
dzbarsky: want to upstream these defaults to windows_support module and bump that?

dzbarsky (approved):

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
```diff
@@ -292,36 +363,76 @@ cc_library(
...
+        # clang-cl's /MD and /MT defaults do not name the Microsoft C++
+        # exception_ptr provider. Make libc++'s archive own that narrow ABI
+        # dependency through a COFF .drectve instead of a Bazel link argument.
+        ":windows_msvc_static_crt": [
+            "/clang:-Xclang",
+            "/clang:--dependent-lib=libcpmt",
+        ],
+        ":windows_msvc": [
+            "/clang:-Xclang",
+            "/clang:--dependent-lib=msvcprt",
```
zbarsky-openai, line 418: **[P1] Do not silently link Microsoft STL into libc++ consumers**

`--dependent-lib=msvcprt` and the `/MT` branch's `--dependent-lib=libcpmt` inject Microsoft's dynamic/static C++ standard-library archives, not just VCRuntime; `/MD` can consequently require `msvcp140.dll`. This directly contradicts the stated libc++-only behavior. The action test misses it because the dependencies live in COFF `.drectve` sections rather than the linker command line. Use a non-STL provider for the required ABI symbols, or explicitly change and verify the supported dependency/runtime behavior.

toolchain/features/msvc/BUILD.bazel
```diff
@@ -0,0 +1,283 @@
...
+)
+
+cc_feature(
+    name = "static_linking_mode",
+    overrides = "@rules_cc//cc/toolchains/features:static_linking_mode",
+)
+
+config_setting(
+    name = "static_crt_config",
+    values = {"features": "static_link_msvcrt"},
```
zbarsky-openai, line 196: **[P1] Honor CRT features selected on individual C/C++ targets**

This `config_setting` sees only configuration-wide `--features`, not `cc_binary(features = ["static_link_msvcrt", "-dynamic_link_msvcrt"])`. Since the CRT features themselves contain no arguments, a target requesting static CRT still receives `/MD` and the dynamic-runtime library directory. The new artifact matrix conceals this by rewriting `//command_line_option:features` in a transition. Make CRT flags and runtime inputs follow the actual selected feature, or reject unsupported target-local selection explicitly.

runtimes/msvc/BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+    "vcruntime_new_debug.h",
+    "vcruntime_string.h",
+    "vcruntime_startup.h",
+    "vcruntime_typeinfo.h",
+]
+
+validated_directory_info(
+    name = "msvc_include",
+    src = "@msvc_runtime//:msvc_include",
+    expected_path_suffix = "/Contents/VC/Tools/MSVC/{}/include".format(MSVC_PAYLOAD_VERSION),
```
zbarsky-openai: **[P1] Respect the consuming root module's SDK/MSVC configuration**

`windows_support` gives a root module's `configure(...)` tag precedence over this dependency's tag, but these checks require the hard-coded versions from `versions.bzl`. A consumer configuring `msvc_version = "14.44.35207"` gets an include directory ending in `/MSVC/14.44.35207/include`, which this rule rejects because it expects `/MSVC/14.50.35717/include`; alternate Windows SDK versions fail the same way. Derive the paths/version data from the selected generated repositories, or isolate this toolchain's repositories, so existing downstream `windows_support` configurations remain usable.

tools/coff_def_parser/def_parser.c
```diff
@@ -0,0 +1,525 @@
...
+    free(contents);
+    return 0;
+  }
+  if (!has_bytes(file_size, 0, COFF_HEADER_SIZE)) {
+    fprintf(stderr, "Object file '%s' is too small for a COFF header\n",
+            filename);
+    free(contents);
+    return 0;
+  }
+  machine = read_u16(contents);
```
zbarsky-openai: **[P1] Make generated DEF exports work with ThinLTO objects**

The generated-DEF action passes each object directly to this parser, but enabling `thin_lto` produces LLVM-bitcode `.obj` files that satisfy neither the COFF nor BIGOBJ branch. Compiling `clang --target=x86_64-pc-windows-msvc -flto=thin -c -x c /dev/null` and passing the resulting object to this PR's parser reproduces `unrecognized file format ... 17218`. Thus the two advertised features work separately but fail together for a DLL with generated exports. Use a bitcode-capable LLVM symbol reader or reject this combination clearly before execution.

zbarsky-openai (commented):
Windows MSVC target support: the inline findings cover nine high-priority correctness or coverage issues and 13 compatibility, maintainability, or simplification issues. The existing failing CI job lost remote-cache inputs and is not attributed to this change.

tools/coff_def_parser/def_parser_main.c
```diff
@@ -0,0 +1,129 @@
...
+    if (argv[index][0] == '@') {
+      FILE *parameters = coff_def_parser_open_file(argv[index] + 1, "rb");
+      char *line;
+      if (parameters == NULL) {
+        fprintf(stderr, "Could not open parameter file: %s\n", argv[index]);
+        coff_def_parser_destroy(parser);
+        fclose(output);
+        return 1;
+      }
+      while ((line = read_line(parameters)) != NULL) {
```
zbarsky-openai: **[P1] Decode the shell-quoted response files that rules_cc actually emits**

The generated-DEF action always creates its response file with `argv.set_param_file_format("shell")`, but this loop only trims each line and passes the still-quoted/escaped text directly to `fopen`. Bazel permits spaces and shell metacharacters in package and target names, so a legal object path can be shell-quoted. Passing a response entry such as `"/tmp/input file.def"` to this PR's compiled parser reproduces `Couldn't open file '"/tmp/input file.def"'`. Parse Bazel's shell-format entries correctly or change the producer to an unquoted format.

toolchain/llvm/llvm.bzl
```diff
@@ -127,6 +180,51 @@ def declare_llvm_targets(*, suffix = ""):
...
     )
 
+    # MSVC ABI actions use the CL/LINK dialect directly. Keep this map separate
+    # from MinGW so target ABI, never execution OS, selects the personality.
+    MSVC_CONSTRUCTION_TOOLS = {
+        "@rules_cc//cc/toolchains/actions:c_compile": ":clang-cl",
+        "@rules_cc//cc/toolchains/actions:cpp_compile": ":clang-cl",
+        "@rules_cc//cc/toolchains/actions:linkstamp_compile": ":clang-cl",
+        "@rules_cc//cc/toolchains/actions:lto_backend": ":clang-cl",
+        "@rules_cc//cc/toolchains/actions:preprocess_assemble": ":clang-cl",
```
zbarsky-openai, line 194: **[P1] Register the plain assembly action as well as preprocessed assembly**

`MSVC_CONSTRUCTION_TOOLS` registers `preprocess_assemble` but omits the separate `assemble` action; the clang-cl input/output action lists omit it too. Consequently `cc_library(srcs = ["foo.s"])` has neither a configured assembler nor its `/c`/`/Fo` argument handling for either MSVC target. The existing fixture only uses `windows_msvc_assembly.S`, which exercises `preprocess_assemble` and cannot detect the failure. Configure both assembly actions or explicitly reject unsupported raw assembly.

toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -0,0 +1,595 @@
...
+    name = "def_file",
+    args = [":def_file_args"],
+    feature_name = "def_file",
+)
+
+# Temporary compatibility boundary. The repository's e2e configuration
+# requests layering_check globally, but CL-mode module-map and header-parsing
+# actions have not yet been implemented. Register the names without enabling
+# module args; do not treat this no-op as support.
+cc_feature(
```
zbarsky-openai, line 626: **[P1] Do not turn requested header-layering checks into silent no-ops**

These empty `layering_check` and `use_module_maps` features make an explicitly requested check appear supported while enforcing nothing. The repository enables `layering_check` globally, and the new action test confirms that strict-decluse, private-header diagnostics, and module-map arguments are all absent. Undeclared/private includes that other toolchains reject therefore compile successfully under MSVC. Fail unsupported enforcement requests, or explicitly disable them in the requesting configuration, until real enforcement exists.

e2e/rules_cc/windows_msvc_artifacts_test.sh
```diff
@@ -0,0 +1,141 @@
...
+found_dll_consumer=0
+found_import_library=0
+found_pdb=0
+debug_executable=
+debug_pdb=
+
+assert_machine() {
+  local artifact="$1"
+  "${LLVM_READOBJ}" --file-headers "${artifact}" |
+    grep -Fq "Machine: IMAGE_FILE_MACHINE_${MACHINE}" ||
```
zbarsky-openai: **[P1] Match the COFF machine exactly so ARM64EC cannot pass as ARM64**

For the ARM64 matrix, `grep -Fq "Machine: IMAGE_FILE_MACHINE_ARM64"` also succeeds on `Machine: IMAGE_FILE_MACHINE_ARM64EC (0xA641)`. An unsupported ARM64EC executable, DLL, or archive therefore passes the supposedly native-ARM64 architecture gate. Match the complete machine token, including its boundary, and verify that ARM64EC is rejected.

.github/workflows/ci.yaml
```diff
@@ -84,6 +84,26 @@ jobs:
...
+            --jobs=100
+            --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7
+          )
+          bash ./windows_msvc_action_test.sh "${common_flags[@]}"
+          bash ./windows_msvc_analysis_test.sh "${common_flags[@]}"
+          bazel --bazelrc=.bazelrc test \
+            "${common_flags[@]}" \
+            --repo_env=BAZEL_MSVC_RUNTIME_VISUAL_STUDIO_EULA=1 \
+            --repo_env=BAZEL_WINDOWS_SDK_EULA=1 \
+            //:windows_msvc_artifacts_matrix
```
zbarsky-openai: **[P1] Actually run the new portable-helper tests in CI**

Root-module CI selects only `//tests/...`, while the e2e jobs run `//...` inside separate workspaces. This step likewise runs only the artifact matrix, so none of the six new unit/CLI targets under `//tools/case_insensitive_copy`, `//tools/case_insensitive_vfs`, and `//tools/coff_def_parser` is executed. Building these helpers as tools does not exercise their existing malformed-input, filesystem, or parser tests. Add the existing helper targets to a root-module CI test invocation.

toolchain/features/BUILD.bazel
```diff
@@ -263,10 +371,13 @@ cc_args(
 
 cc_feature(
     name = "generate_pdb_file",
-    args = [
-        ":generate_pdb_file_compile_args",
-        ":generate_pdb_file_link_args",
-    ],
+    args = select({
+        "//constraints/windows/abi:msvc": ["//toolchain/features/msvc:pdb_args"],
```
zbarsky-openai, line 444: **[P2] Include compile-time CodeView data when generating a PDB**

The existing `generate_pdb_file` feature enables both compilation debug records and linker PDB generation, but this MSVC branch replaces both with `pdb_args`, which only passes linker `/DEBUG`. Under fastbuild/opt, `cc_binary(features = ["generate_pdb_file"])` therefore creates a valid-looking PDB without source/type records because its objects were not compiled with `/Z7`, `/Zi`, or an equivalent option. The new debug fixture checks only the PDB file and reference. Preserve both compile and link portions of the feature.

toolchain/features/msvc/validation.bzl
```diff
@@ -0,0 +1,73 @@
...
+    if not is_msvc_abi:
+        return None
+    if is_legacy_msvcrt:
+        return "MSVC ABI requires //constraints/windows/crt:ucrt; legacy msvcrt is MinGW-only"
+    if not is_libcxx:
+        return "Layer 1 MSVC ABI requires //constraints/cxxstdlib:libcxx"
+
+    enabled = {feature: True for feature in enabled_features}
+    disabled = {feature: True for feature in disabled_features}
+    if "dynamic_link_msvcrt" in disabled and "static_link_msvcrt" not in enabled:
```
zbarsky-openai, line 33: **[P2] Reject simultaneously enabled dynamic and static CRT modes**

This condition rejects a missing CRT selection but accepts `--features=dynamic_link_msvcrt,static_link_msvcrt`, despite claiming to require exactly one mode. Both feature names then remain enabled while `static_crt_config` independently selects `/MT`, so downstream feature consumers can observe a runtime different from the emitted flags and library inputs. Make the two modes mutually exclusive and reject both-enabled as well as neither-enabled configurations.

toolchain/features/legacy/BUILD.bazel
```diff
@@ -5,45 +5,118 @@ load("@rules_cc//cc/toolchains:nested_args.bzl", "cc_nested_args")
...
+    name = "clang_cl_legacy_default_frame_pointer",
+    actions = ["@rules_cc//cc/toolchains/actions:compile_actions"],
+    args = ["/clang:-fno-omit-frame-pointer"],
+)
+
+cc_feature(
+    name = "clang_cl_default_compile_flags",
+    args = [
+        "//toolchain/args:clang_cl_suppress_compiler_banner",
+        "//toolchain/args:clang_cl_default_exception_handling",
```
zbarsky-openai, line 131: **[P2] Preserve the existing release configuration under clang-cl**

The repository's existing `--config=release` supplies `-fno-exceptions`, `-fno-rtti`, and `-fomit-frame-pointer`, while this new personality enables `/EHsc`, `/GR`, and preserved frame pointers. A direct `clang --driver-mode=cl ... -###` run reports all three GNU-form release options as ignored and still enables C++ exceptions. MSVC release builds therefore silently violate the existing exception/RTTI/frame-pointer policy. Translate these options for clang-cl or reject the incompatible release configuration.

tools/coff_def_parser/def_parser.c
```diff
@@ -0,0 +1,525 @@
...
+    uint32_t symbol_count = read_u32(contents + 12);
+    result = parse_symbol_table(parser, contents, file_size, section_offset,
+                                section_count, symbol_offset, symbol_count,
+                                COFF_SYMBOL_SIZE, 0, arch_for_machine(machine),
+                                filename);
+  } else if (has_bytes(file_size, 0, BIGOBJ_HEADER_SIZE) &&
+             read_u16(contents) == 0 && read_u16(contents + 2) == 0xffff) {
+    machine = read_u16(contents + 6);
+    result =
+        parse_symbol_table(parser, contents, file_size, BIGOBJ_HEADER_SIZE,
```
zbarsky-openai: **[P2] Validate the machine type in the BIGOBJ branch**

Ordinary COFF objects must satisfy `supported_machine(machine)`, but BIGOBJ files bypass that check after reading their machine field. A structurally valid BIGOBJ with unsupported machine `0x1234` currently succeeds and exports its `foreign` symbol instead of being rejected. Apply the same machine validation before parsing BIGOBJ symbol tables.

runtimes/msvc/BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+        ":msvc_com_support_headers",
+        ":msvc_vcruntime_headers",
+        ":winsdk_shared_include",
+        ":winsdk_ucrt_include",
+        ":winsdk_um_include",
+        ":winsdk_winrt_include",
+    ],
+)
+
+case_insensitive_copy_directory(
```
zbarsky-openai: **[P2] Delete the redundant SDK library-copy actions and helper**

`MODULE.bazel` already configures lowercase aliases for `.lib`/`.Lib`, and `windows_support` creates those aliases before constructing the validated SDK `DirectoryInfo` trees. `:winsdk_ucrt_lib` and `:winsdk_um_lib` can therefore be passed directly to the linker search-path arguments. Remove these two full-directory copies and the dedicated copy rule/tool/tests—about 488 added lines, plus unnecessary per-architecture actions and inputs. This also removes a helper that follows directory symlinks even though the VFS helper correctly rejects them.

MODULE.bazel
```diff
@@ -105,6 +106,42 @@ use_repo(musl, "musl_libc")
...
+        "base/c/Include/10.0.26100.0/shared/driverspecs.h": "base/c/Include/10.0.26100.0/shared/DriverSpecs.h",
+        "base/c/Include/10.0.26100.0/shared/specstrings.h": "base/c/Include/10.0.26100.0/shared/SpecStrings.h",
+        "base/c/Include/10.0.26100.0/um/ole2.h": "base/c/Include/10.0.26100.0/um/Ole2.h",
+        "base/c/Include/10.0.26100.0/um/olectl.h": "base/c/Include/10.0.26100.0/um/OleCtl.h",
+        "**/*.h": "lowercase",
```
zbarsky-openai: **[P2] Remove SDK header transformations already handled by the VFS**

The new Clang VFS overlay is explicitly case-insensitive, so `"**/*.h": "lowercase"` and the four special-case header aliases duplicate behavior already provided during compilation. Generating the actual new overlay for a directory containing only `Windows.h` allows `#include <wInDoWs.H>` without any aliases. `windows_support` scans SDK files and creates symlinks for these transformations, unnecessarily expanding the repository/action inputs. Delete the header transformations and retain only the `.lib` normalization required by the linker.

toolchain/cc_toolchain.bzl
```diff
@@ -1,9 +1,92 @@
...
+_WINDOWS_MSVC_SUPPORTS_HEADER_PARSING = False
+
+def cc_toolchain(
+        name,
+        tool_map,
+        module_map = None,
+        extra_args = None):
+    extra_args = extra_args or []
+    cc_feature_set(
+        name = name + "_msvc_known_features",
```
zbarsky-openai, line 13: **[P2] Stop creating MSVC feature groups for every toolchain instance**

This macro now creates 15 per-instance feature-set/alias targets instead of three. Across six prebuilt execution toolchains and 27 source-bootstrap toolchains, that adds 396 synthetic targets to the globally evaluated package; all 27 bootstrap instances construct MSVC groups even though MSVC bootstrap toolchains are deliberately unsupported. Hoist shared feature groups, keep direct selections where possible, and instantiate MSVC-only structures only for eligible Linux-hosted prebuilt toolchains.

toolchain/BUILD.bazel
```diff
@@ -384,7 +389,78 @@ cc_args_list(
...
+        "//toolchain/args:ignore_unused_command_line_argument",
+    ],
+)
+
+# Keep the complete Windows composition explicit while its individual
+# compiler-personality, object-format, ABI, and runtime semantics are being
+# normalized. This mirrors :generic_toolchain_args instead of appending a bag
+# of Windows-only arguments through :platform_specific_args.
+cc_args_list(
+    name = "windows_toolchain_args",
```
zbarsky-openai, line 402: **[P2] Remove the second near-identical ordered toolchain composition**

This Windows argument list repeats 33 of 37 generic entries, and `toolchain/runtimes/BUILD.bazel` repeats another 27 of 31 in its own Windows variant. Sanitizer and link-order changes now have to remain synchronized across four separate lists. Factor the shared ordered groups once and select only the few Windows replacements/insertions at their existing positions; this removes roughly a hundred duplicated entries and prevents configuration drift.

defs.bzl
```diff
@@ -1,8 +1,18 @@
 def exec_test(*, rule, name, tags = [], args = [], env = {}, data = [], tools = [], **kwargs):
+    # The inner executable is built in the exec configuration. Keep target
+    # artifacts on the outer test unless an inner rule attribute itself uses a
+    # location expansion for one of them (for example go_test.x_defs).
+    inner_data = []
+    if type(data) == "list":
+        inner_data = [
+            item
+            for item in data
+            if " {}".format(item) + ")" in str(kwargs)
```
zbarsky-openai: **[P2] Do not infer declared prerequisites by searching `str(kwargs)`**

The existing wasm caller works only because its `data` label and `x_defs` location expression use identical spelling. Replace either spelling with the equivalent `wasm_blob` or `//tests/wasm:wasm_blob` and this substring check drops the inner `go_test` prerequisite, causing location expansion to fail; any configurable `data = select(...)` is also discarded entirely. Replace this brittle shared-macro heuristic with an explicit `inner_data` argument or another label-aware ownership boundary.

e2e/rules_cc/BUILD.bazel
```diff
@@ -327,6 +333,90 @@ cc_binary(
...
+
+[
+    cc_test(
+        name = "windows_msvc_libcxx_behavior_" + crt,
+        srcs = [
+            "windows_msvc_assembly.S",
+            "windows_msvc_c_smoke.c",
+            "windows_msvc_libcxx_behavior.cc",
+        ],
+        copts = ["/std:c++20"],
```
zbarsky-openai: **[P2] Delete C++-only flags from mixed C/assembly fixtures**

This `/std:c++20` option is applied to `.cc`, `.c`, `.S`, and ThinLTO backend actions. Both exact-head Windows x64 and ARM64 CI logs contain six actual `argument unused during compilation: '/std:c++20'` warnings, which become failures for warnings-as-errors users. The fixtures need only the existing C++17 default, so remove the unnecessary `/std:c++20` overrides instead of suppressing the warning.

platforms/declare_platforms.bzl
```diff
@@ -45,6 +45,21 @@ def declare_platforms():
             )
 
     declare_platforms_libc_aware()
+    declare_windows_msvc_platforms()
+
+def declare_windows_msvc_platforms():
+    for target_cpu in ["x86_64", "aarch64"]:
+        native.platform(
+            name = "windows_{}_msvc".format(target_cpu),
```
zbarsky-openai, line 53: **[P2] Update the user-facing Windows support documentation**

The unchanged root `README.md` still explicitly says MSVC targets have no compatible C/C++ toolchain, and its supported-platform matrix omits both new MSVC platforms. Document `windows_x86_64_msvc` and `windows_aarch64_msvc`, their Linux-only execution requirement, the two Microsoft EULA repository flags, static libc++, and the retail-only `/MD`/`/MT` limitation; `/MDd` and `/MTd` are not implemented despite the PR description mentioning debug CRT selection. Otherwise users are told this new public feature does not exist.

extensions/llvm.bzl
```diff
@@ -38,6 +38,7 @@ _LLVM_21_SOURCE_PATCHES = [
     "//3rd_party/llvm-project/21.x/patches:no_rules_python.patch",
     "//3rd_party/llvm-project/21.x/patches:llvm-windows-stack-size.patch",
     "//3rd_party/llvm-project/21.x/patches:libcxx-lgamma_r.patch",
+    "//3rd_party/llvm-project/21.x/patches:libcxx-vcruntime-nothrow.patch",
```
zbarsky-openai: **[P2] Keep the identical libc++ patch in one shared location**

The newly added 21.x, 22.x, and x.x `libcxx-vcruntime-nothrow.patch` files are byte-for-byte identical, and every supported LLVM major already inherits `_DEFAULT_SOURCE_PATCHES`. Put the x.x patch in that shared list and delete the two duplicate patch files plus the three per-major registrations.

tools/case_insensitive_vfs/filesystem.c
```diff
@@ -0,0 +1,430 @@
...
+  return result;
+}
+
+static int compare_entries(const void *left, const void *right) {
+  const struct ci_directory_entry *left_entry = left;
+  const struct ci_directory_entry *right_entry = right;
+  return strcmp(left_entry->folded_name, right_entry->folded_name);
+}
+
+static int add_entry(struct ci_directory_entries *entries, const char *name,
```
zbarsky-openai: **[P2] Deduplicate SDK directory entries after sorting**

`add_entry` scans every previous folded name and reallocates the array on each insertion, then the completed array is sorted anyway. A directory with 6,000 SDK entries therefore performs roughly 18 million case-name comparisons before `qsort`, and generated header aliases make that worse. Append with geometric growth, sort once by folded name, and collapse adjacent aliases/collisions to preserve deterministic output without the quadratic scan.

MODULE.bazel
```diff
@@ -105,6 +106,42 @@ use_repo(musl, "musl_libc")
...
+        "**/*.lib": "lowercase",
+        "**/*.Lib": "lowercase",
+    },
```
ArchangelX360: Is it still needed given we have `ivfsoverlay` support now?

3rd_party/llvm-project/21.x/patches/libcxx-vcruntime-nothrow.patch
cerisier: Yes, it's upstreamed already.

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
cerisier: Will have a look

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
cerisier: I'm ok with verbosity, fully agree.

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
cerisier: Worth investigating actually, we don't use `static_runtime_lib` for libc++.lib on msvc, we discover it using search paths. I have to see if this is limitation of clang or if it's fully wrong here.

e2e/rules_cc/windows_msvc_action_test.sh
cerisier: I have a follow up TODO to transfer that into analysis tests. If possible in nother PR.

runtimes/msvc/BUILD.bazel
cerisier: Sure.

runtimes/msvc/versions.bzl
cerisier: Yes it should. Let's do that.

tests/linux_linking/BUILD.bazel
cerisier: Yes, we'll do that.

toolchain/features/clang_cl/BUILD.bazel
cerisier: Were crafter and reviewed to be upstreamable. The separation is clear, clang_cl vs msvc, especially for that reason.

toolchain/declare_toolchains.bzl
cerisier: Sure !

MODULE.bazel
cerisier: I'm not sure just yet because I MAY hve a way to get rid of that entirely.

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
cerisier: I understand and will check, but what's the solution here, what can I use instead of /MD if it brings implicit deps...

toolchain/features/msvc/BUILD.bazel
cerisier: Didn't undedstand the proposed solution

runtimes/msvc/BUILD.bazel
cerisier: Probably fixed when we generate the versions from windows_support.

tools/coff_def_parser/def_parser.c
cerisier: Need to validate the actual scenario and if it matters.

tools/coff_def_parser/def_parser_main.c
cerisier: Good catch. we should add a test for that.

toolchain/llvm/llvm.bzl
cerisier: Seems valid, let me double check if this was intentional.

toolchain/features/clang_cl/BUILD.bazel
cerisier: Makes me think we might as well try to implement boths.

e2e/rules_cc/windows_msvc_artifacts_test.sh
cerisier: Easy one, agreed.

.github/workflows/ci.yaml
cerisier: Agreed.

toolchain/features/BUILD.bazel
cerisier: Do you mean we should imply "dbg" ?

toolchain/features/msvc/validation.bzl
cerisier: Valid.

toolchain/features/legacy/BUILD.bazel
cerisier: Right, how do we approach that ? Do we use transitions instead of copts ?

tools/coff_def_parser/def_parser.c
cerisier: IF we can have a failing test, then ok.

runtimes/msvc/BUILD.bazel
cerisier: If we can validate with compiling clang for windows MSVC from mac with RBE linux, then this is my favorite change so far.

toolchain/cc_toolchain.bzl
cerisier: How would you structure that ? low prio.

toolchain/BUILD.bazel
cerisier: Ok

defs.bzl
cerisier: Make a proposal.

e2e/rules_cc/BUILD.bazel
cerisier: Ok

platforms/declare_platforms.bzl
cerisier: Ok

extensions/llvm.bzl
cerisier: No, because in LLVM 32 it's landed I think. if not, ok.

tools/case_insensitive_vfs/filesystem.c
cerisier: Ok

MODULE.bazel
cerisier: So it's more complicated but they should go indeed.

vfs is compile time only. so all of .lib is not covered.
However, we use case insensitive copy so the transformations here can go.

But, I wonder if we could find a way to make a vfs for link as well and get rid of all.
I'll give it more thoughts.

runtimes/msvc/BUILD.bazel
cerisier: Thinking more about this, I think this is not correct because windows_support makes the assumption that the host fs is going to be the exec fs, so sadly something is wrong here

3rd_party/llvm-project/x.x/compiler-rt/compiler-rt.BUILD.bazel
cerisier: Checked the current tree: these predicates are consumed only within compiler-rt, so moving them to //private/constraints would create a shared abstraction without a shared consumer. I am keeping them local and simplifying the redundant CPU-specific composition in the adjacent change.

tools/coff_def_parser/def_parser_main.c
cerisier: Fixed in caf39b3c. The parser now decodes each shell-format response-file entry before opening it, including quoted paths and backslash escapes, and rejects malformed or multi-argument entries. The quoted-space regression failed before (BuildBuddy 4e4268f6-7fc2-4006-a65b-79ffa3be6efe) and passes after (40f49741-ed27-4461-b5aa-654b4b1f2e19).

toolchain/llvm/llvm.bzl
cerisier: Fixed in a5e1159d. The clang-cl tool map now registers the plain assemble action, with assemble-specific input/output arguments, and the e2e fixture includes real x86-64 and ARM64 .s sources. Baseline analysis failed with no assemble action; both target builds pass after (x64 f65585ad-2d1a-4820-a772-75a174ff4531, ARM64 8ec3fb89-cb31-4ef7-9d76-7bf26d61a83c).

toolchain/features/msvc/validation.bzl
cerisier: Fixed in 1918a1f1. dynamic_link_msvcrt and static_link_msvcrt now provide one mutually exclusive crt_mode category, so feature configuration rejects both-enabled requests while the existing validation still rejects neither-enabled requests. The focused analysis suite passes; the conflicting request fails at feature configuration (BuildBuddy 7b97d58a-b949-4e66-83df-6b711a702eca).

platforms/declare_platforms.bzl
cerisier: Fixed in d3eed610. The README now lists both MSVC target platforms and their Linux-only execution support, documents both EULA flags, static libc++, the /MD default and explicit /MT selection, and the unsupported /MDd and /MTd modes. Platform-label queries and the Gazelle diff check pass (BuildBuddy d1ad09fa-0e2c-4d2a-9824-54fba5d33b18).

toolchain/BUILD.bazel
cerisier: Deferring this composition refactor for now. The current ordered Windows list is intentional pending a broader semantic-group consolidation; changing it in this PR would couple structural cleanup to the remaining correctness fixes.

extensions/llvm.bzl
cerisier: Not changing patch placement in this PR. The supported LLVM lines do not all share the same upstream boundary, so we will keep the explicit version registrations until that version matrix is handled separately.

zbarsky-openai (commented):
Fresh review of `2eb23e5` after reading Corentin's responses and validating the current CI/artifacts. Three reproduced high-priority issues remain: transitive ThinLTO objects bypass generated-DEF validation, raw assembly actions emit dozens of invalid frontend/SDK arguments, and SDK overlays write nonexistent header paths into depfiles. The inline comments also cover the mixed-architecture archive bypass and concrete reductions: reuse shipped `llvm-nm`, collapse duplicate SDK/runtime copies and compiler-policy features, remove dead parser/VFS code and unrelated test wrappers, and drop or verify untested macOS execution scope. The PR description should also accurately describe upstream's required Microsoft STL runtime and the unsupported debug CRT.

toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -0,0 +1,657 @@
...
+        ":thin_lto_object_suffix_replace_args",
+        ":thin_lto_merged_object_args",
+        ":thin_lto_backend_compile_args",
+        ":thin_lto_backend_index_args",
+        ":thin_lto_backend_output_args",
+        ":thin_lto_backend_input_args",
+    ],
+    feature_name = "thin_lto",
+    implies = [":supports_start_end_lib"],
+    mutually_exclusive = ["//toolchain/features:generated_def_object_format"],
```
zbarsky-openai: **[P1] Check ThinLTO across generated-DEF dependencies, not just the DLL**

This new exclusion observes only the DLL target's features. With global `--features=thin_lto`—including the repository's release configuration—`cc_binary(linkshared = True, features = ["windows_export_all_symbols", "-thin_lto"], deps = [":lib"])` passes this check, but `:lib` still emits bitcode. `rules_cc` explicitly adds transitive library objects to the generated-DEF response file, and the current parser then fails on those objects. Enforce one graph-wide object format or use the existing bitcode-capable `llvm-nm`; the existing negative case checks only a same-target conflict.

toolchain/args/BUILD.bazel
```diff
@@ -33,14 +33,48 @@ cc_args(
...
+
+cc_args(
+    name = "clang_cl_suppress_compiler_banner",
+    actions = ["@rules_cc//cc/toolchains/actions:source_compile_actions"],
+    args = ["/nologo"],
+)
+
+cc_args(
+    name = "clang_cl_default_exception_handling",
+    actions = ["@rules_cc//cc/toolchains/actions:source_compile_actions"],
```
zbarsky-openai: **[P1] Keep frontend and SDK arguments off raw assembly actions**

`source_compile_actions` includes raw `assemble`, so the newly supported `.s` actions receive `/EHsc`, `/GR`, all SDK `/D` and `/imsvc` arguments, the VFS overlay, and CodeView settings. Both exact-head native Windows CI jobs emit **44 real unused-argument warnings each**; compiling the checked-in ARM64 `.s` with `-Werror=unused-command-line-argument` fails immediately. Scope C++ arguments to C++ frontend actions and SDK/preprocessor/debug arguments to actual frontend/preprocessed-assembly actions; raw assembly should receive only its applicable target/input/output settings.

runtimes/msvc/BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+    root_paths = ["**/include"],
+)
+
+directory_info(
+    name = "msvc_com_support_headers",
+    src = ":msvc_com_support_headers_source",
+)
+
+case_insensitive_vfs_overlay(
+    name = "msvc_sdk_header_case_overlay",
```
zbarsky-openai, line 127: **[P1] Stop emitting nonexistent SDK header paths into compiler depfiles**

This overlay explicitly disables LLVM's default external-name behavior even though clang-cl always emits `-MD`. With physical `SDK/Nested/CaseHeader.H` and `#include "nested/caseheader.h"`, I reproduced a `.d` entry for nonexistent `SDK/nested/caseheader.h`; using external names records the real declared header. Bazel's header discovery resolves paths case-sensitively on Linux/macOS and rejects unresolved entries as undeclared inclusions, and the removed SDK aliases contain these exact case mismatches. LLVM already defaults `use-external-names` to true: omit this JSON property for both overlays and delete the now-unnecessary boolean, CLI switch, and dual-mode tests.

e2e/rules_cc/windows_msvc_artifacts_test.sh
```diff
@@ -0,0 +1,151 @@
...
+debug_pdb=
+
+machine_line_matches() {
+  local machine="$1"
+  grep -Eq "Machine: IMAGE_FILE_MACHINE_${machine}([[:space:]]|$)"
+}
+
+assert_machine() {
+  local artifact="$1"
+  "${LLVM_READOBJ}" --file-headers "${artifact}" |
```
zbarsky-openai, line 50: **[P2] Validate every COFF archive member's machine**

The tighter ARM64 regex still accepts a mixed-architecture archive because this pipeline succeeds when **any** member matches. I created a real `.lib` containing one ARM64 object and one ARM64EC object; `llvm-readobj --file-headers` reports both machines, and the current ARM64 assertion exits successfully. Require every reported `Machine:` line to equal the expected architecture, while also requiring at least one header, so unsupported ARM64EC members cannot hide behind a valid member.

platforms/common.bzl
```diff
@@ -31,6 +31,15 @@ SUPPORTED_EXECS = [
     ("windows", "aarch64"),
 ]
 
+# MSVC target toolchains use the published Unix execution compilers. Native
+# Windows execution remains gated until its compiler is selected and proved.
+MSVC_TARGET_SUPPORTED_EXECS = [
+    ("macos", "x86_64"),
```
zbarsky-openai, line 37: **[P2] Remove or actually verify the newly added macOS execution scope**

These entries advertise four additional macOS-executor/MSVC-target combinations, although the PR description scopes compiler/linker execution to Linux. I checked all 38 exact-head CI jobs: both macOS runners build only native macOS targets/sanitizers, and neither runs a single MSVC-target action. The existing MSVC action checks even hard-code Linux toolchain paths. The simplest change is to drop these unverified macOS registrations and README checkmarks; if macOS execution is intended, run the existing MSVC smoke build on both existing macOS jobs before claiming support.

toolchain/llvm/llvm.bzl
```diff
@@ -99,6 +147,11 @@ def declare_llvm_targets(*, suffix = ""):
         },
     )
 
+    cc_tool(
+        name = "coff_def_parser",
```
zbarsky-openai: **[P2] Replace the handwritten COFF parser with the already-shipped LLVM symbol reader**

This tool introduces roughly **1,270 lines** of COFF/BIGOBJ parsing, CLI/tests, and licensing even though the adjacent `static_library_validator` already declares hermetic `llvm-nm`. I verified `llvm-nm --defined-only --extern-only --format=posix` reads both ordinary Windows COFF and ThinLTO objects and exposes function/data symbol types. `rules_cc` invokes this action only as `<tool> <output.def> <dllname> @object-rsp`; explicit `.def` files bypass it entirely. Reuse that existing `LLVM_NM` tool/data pattern with a small response-file/DEF-format adapter, preserve required symbol/DATA filtering, and delete the custom binary parser, unreachable `.def` merging, and ThinLTO exclusion.

tools/coff_def_parser/def_parser.c
```diff
@@ -0,0 +1,531 @@
...
+  size_t index;
+  for (index = 0; index < set->count; ++index) {
+    free(set->values[index]);
+  }
+  free(set->values);
+}
+
+static void string_set_insert(struct string_set *set, const char *value) {
+  size_t position = 0;
+  while (position < set->count && strcmp(set->values[position], value) < 0) {
```
zbarsky-openai: **[P2] Avoid quadratic export-set insertion if this parser remains**

Each symbol insertion scans the already-sorted vector and may shift its tail. Using real valid AMD64 COFF objects, this helper took 0.040s, 0.156s, 0.612s, and 2.35s for 4k, 8k, 16k, and 32k exports—almost **4× slower per doubling**. Large generated DLLs therefore pay quadratic parsing cost. If the parser is retained instead of using `llvm-nm`, append symbols, sort once before writing, and skip adjacent duplicates; that is both smaller and asymptotically faster.

runtimes/msvc/BUILD.bazel
```diff
@@ -0,0 +1,211 @@
...
+    expected_path_suffix = "/Contents/VC/Tools/MSVC/%s/lib/arm64" % MSVC_PAYLOAD_VERSION,
+)
+
+validated_directory_info(
+    name = "msvc_lib_x64",
+    src = "@msvc_runtime//:msvc_lib_x64",
+    expected_path_suffix = "/Contents/VC/Tools/MSVC/%s/lib/x64" % MSVC_PAYLOAD_VERSION,
+)
+
+copy_to_directory(
```
zbarsky-openai, line 100: **[P2] Collapse the duplicated MSVC header and runtime-copy trees**

The COM and VCRuntime header copies traverse the same `:msvc_include` and have disjoint allowlists; one combined copy/`DirectoryInfo`/VFS root/`/imsvc` preserves the exact visible headers. Likewise, copy `comsuppw.lib` together with the configuration-selected static or dynamic CRT allowlist into one architecture-selected directory and expose one `/LIBPATH`. This removes three copy-rule definitions, two executed copy actions, repeated payload traversal, one include path, and one linker path from every MSVC build without exposing Microsoft STL headers.

toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -0,0 +1,657 @@
...
+
+_LINK_ACTIONS = ["@rules_cc//cc/toolchains/actions:link_actions"]
+
+_LTO_INDEX_ACTIONS = [
+    "@rules_cc//cc/toolchains/actions:lto_index_for_dynamic_library",
+    "@rules_cc//cc/toolchains/actions:lto_index_for_executable",
+    "@rules_cc//cc/toolchains/actions:lto_index_for_nodeps_dynamic_library",
+]
+
+cc_args(
```
zbarsky-openai, line 38: **[P2] Define release-policy features once across compiler personalities**

`no_exceptions`, `no_rtti`, `omit_frame_pointer`, their argument targets, and their grouping are now duplicated here and in `//toolchain/features:BUILD.bazel`. `cc_args.actions` and `args` both support `select()`, so define each semantic feature once with ABI-selected action sets and spellings (`/EHs-c-` versus `-fno-exceptions`, etc.), and reference the same shared `compiler_policy_features` from both toolchains. This preserves the narrower clang-cl C++/ThinLTO action scopes while removing seven duplicate targets and roughly 50 lines.

tools/case_insensitive_vfs/filesystem.c
```diff
@@ -0,0 +1,388 @@
...
+  size_t size = strlen(value);
+  char *result = ci_xmalloc(size + 1);
+  for (index = 0; index < size; ++index) {
+    result[index] = (char)tolower((unsigned char)value[index]);
+  }
+  result[size] = '\0';
+  return result;
+}
+
+int ci_preferred_name(const char *left, const char *right, char **preferred,
```
zbarsky-openai: **[P2] Delete the now-dead preferred-name implementation and its tests**

After the sort/deduplicate rewrite, `ci_preferred_name` has zero production callers: the only references are its public declaration, a test-only `PreferredName` wrapper, and two tests that exercise this otherwise unreachable function. Actual collision selection now happens independently in `validate_sorted_entries`/`collapse_sorted_entries`, which already have behavioral overlay coverage. Delete the dead function, declaration, wrapper, and the two implementation-only tests.

tests/linux_linking/BUILD.bazel
```diff
@@ -71,7 +72,8 @@ platform_transition_binary(
     target_platform = "@llvm//platforms:linux_aarch64",
 )
 
-sh_test(
+exec_test(
```
zbarsky-openai, line 75: **[P2] Revert the unrelated Linux test-wrapper migration**

All six existing Linux ELF/linking tests worked as ordinary `sh_test` targets before this Windows MSVC change. Wrapping each in `exec_test` adds six extra inner targets, execution transitions, an additional shared-macro dependency, and `data`/`tools` plumbing without exercising an MSVC behavior. Restore this file to its merge-base version unless a concrete Windows-support dependency requires the conversion; this removes unrelated build-graph and review surface.

toolchain/features/msvc/BUILD.bazel
zbarsky-openai: You're right to question this. I retract the target-local CRT suggestion: Bazel does not propagate a target's `features` into transitive libraries or libc++, so selecting `/MT` there could mix `/MT` and `/MD` objects. The configuration-wide `--features=-dynamic_link_msvcrt,static_link_msvcrt` boundary and your graph-wide matrix transition are the coherent design. Documenting that target-local `features` cannot select CRT is sufficient.

3rd_party/llvm-project/x.x/libcxx/libcxx.BUILD.bazel
zbarsky-openai: `/MD` itself is not the problem. Upstream LLVM's `libcxx/CMakeLists.txt` explicitly selects `msvcprt` or `libcpmt` because `exception_ptr` internals require Microsoft STL symbols, and Microsoft's `stl/src/excptptr.cpp` implements the `__ExceptionPtr*` functions. My suggestion to replace those with a nonexistent VCRuntime-only provider was incorrect. Keep the upstream libraries; instead update the PR's repeated “Microsoft STL is not introduced” claim and explicitly document/check the `/MD` `MSVCP140.dll` redistributable dependency in the existing artifact test.

runtimes/msvc/BUILD.bazel
zbarsky-openai: This is still unfixed at `2eb23e5`: `windows_support` gives the root module's configure tag precedence, while these wrappers still require the hard-coded MSVC `14.50.35717` / SDK `10.0.26100.0` paths and the compiler flag still hard-codes `19.50.35717`. A root selecting another supported payload therefore fails analysis or gets the wrong compatibility setting. Consume the upstream `DirectoryInfo` targets directly and export the actual selected compatibility version; that also lets you delete `versions.bzl` and the suffix-validation layer.

toolchain/features/BUILD.bazel
zbarsky-openai: No—do not imply `dbg`, because that also changes optimization and unrelated compilation policy. `generate_pdb_file` should independently emit CodeView at compile time (`/Z7`) and `/DEBUG` at link time, including opt/fastbuild. Reuse the existing `generate_pdb_file_compile_args`, select `/Z7` for MSVC versus `-gcodeview` otherwise, and include that shared compile half alongside the existing MSVC `pdb_args`.

toolchain/features/clang_cl/BUILD.bazel
zbarsky-openai: I would keep this PR smaller: delete `pending_layering_check` and `pending_use_module_maps`, add those names to the existing MSVC unsupported-feature validation, and explicitly disable the globally requested layering feature only in MSVC-specific configuration. Generic toolchains keep real enforcement, and an explicitly re-enabled MSVC check fails instead of silently succeeding. Actual module/header-parsing implementation can then be a separate focused change.

toolchain/cc_toolchain.bzl
zbarsky-openai: None of the 11 `cc_feature_set` targets or four runtime aliases depends on the toolchain's `name`, `tool_map`, or execution CPU. Define the generic/MSVC known, enabled, runtime-enabled, and stage-selected groups once in `//toolchain/features`, pass configurable shared labels directly to `_cc_toolchain(known_features = ..., enabled_features = ...)`, and select runtime-library labels directly as before. That preserves feature ordering but removes all 15 synthetic targets per invocation, including the MSVC groups currently created for unsupported bootstrap toolchains.

toolchain/BUILD.bazel
zbarsky-openai: This can be narrower than a broader semantic-group refactor: in each BUILD file, keep one existing ordered argument list and replace only the differing positions with platform-selected wrappers—the target/deterministic entries, the platform-compile slot, and the final CRT/link tail. The shared sanitizer and linking sequences remain in exactly the same order. Deleting the second Windows list in both files removes roughly 100 duplicated entries without changing action ordering.

extensions/llvm.bzl
zbarsky-openai: Agreed that putting this in `_DEFAULT_SOURCE_PATCHES` would incorrectly cross an upstream version boundary. Keep the three existing major-specific registrations exactly as they are, but point the 21.x and 22.x entries at the same existing `//3rd_party/llvm-project/x.x/patches:libcxx-vcruntime-nothrow.patch` already used by 23.x. All three blobs are byte-identical, so deleting the two duplicate files preserves the exact version gating.

toolchain/features/clang_cl/BUILD.bazel
cerisier: Don't care. Those features are not supposed to be used directly on targets. Old toolchain doesn't support it either.

e2e/rules_cc/windows_msvc_artifacts_test.sh
cerisier: Yeah ok

platforms/common.bzl
cerisier: Only if we did so with mingw

toolchain/llvm/llvm.bzl
cerisier: That would be sick !

tools/coff_def_parser/def_parser.c
cerisier: Can be allegedly ignore by prior command

runtimes/msvc/BUILD.bazel
cerisier: To investigate

toolchain/features/clang_cl/BUILD.bazel
cerisier: To be investigate

tools/case_insensitive_vfs/filesystem.c
cerisier: Ok if true

tests/linux_linking/BUILD.bazel
cerisier: Fix only if it will be forever incorrect.

PR #726, opened 2026-08-30
Add Windows MSVC LLVM prebuilts

cerisier:
## Motivation

Windows LLVM prebuilts currently contain only the GNU/MinGW variant. The Windows MSVC target support merged previously can build programs with clang-cl, libc++, the Microsoft SDK, and UCRT, but the release pipeline cannot yet produce an equivalent optimized MSVC LLVM toolchain.

This change adds separately named MSVC prebuilts while retaining the existing GNU/MinGW Windows archives.

## Invariant and contract

Windows MSVC LLVM release products:

- are built hermetically on Linux remote execution platforms;
- target native AMD64 or ARM64 COFF;
- use clang-cl, llvm-ar, and lld-link;
- use static libc++;
- use ThinLTO and bootstrap FDO;
- use the static MSVC runtime (`/MT`) for release portability;
- do not broaden the public Stage 0 FDO contract;
- do not introduce a native Windows execution toolchain.

Existing GNU/MinGW and non-Windows release behavior remains unchanged.

## Root cause

The target-side MSVC toolchain existed, but the producer graph was incomplete:

- source-backed Stage 2 and Stage 3 MSVC toolchains were not registered;
- bootstrap FDO workloads and profiles did not cover clang-cl MSVC targets;
- profile-use arguments were not expressed using clang-cl syntax;
- supported LLVM release lines lacked required Windows bootstrap patches;
- the release package and workflow exposed only the existing GNU Windows archives.

## Change

- Restore the source-backed Windows MSVC bootstrap prerequisites for supported LLVM versions.
- Generate explicit x64 and ARM64 clang-cl workloads on both Linux executor architectures.
- Merge separate MSVC bootstrap profiles with the declared source-built `llvm-profdata`.
- Apply those profiles to direct Stage 3 ThinLTO builds.
- Keep runtime builds isolated from outer release feature disables.
- Make `--config=release` select `/MT` for MSVC products without changing ordinary consumer `/MD` defaults.
- Add x64 and ARM64 MSVC release archive targets.
- Publish the new archives with `-msvc` suffixes while preserving the existing GNU archive names.

The resulting Windows assets are distinct:

- `windows-amd64` and `windows-arm64` remain GNU/MinGW;
- `windows-amd64-msvc` and `windows-arm64-msvc` are the new MSVC variants.

## Tests

- Built all four Windows release archives together from macOS through Linux RBE: [BuildBuddy invocation](https://app.buildbuddy.io/invocation/3cf0b581-d712-468a-9ac6-e8fc23564b05).
- Built both complete MSVC Stage 3 release archives with ThinLTO and FDO: [BuildBuddy invocation](https://app.buildbuddy.io/invocation/da04b3cf-e3f0-400a-aab6-b4d81bdc22d4).
- Verified all four packaged `llvm.exe` files report their expected AMD64 or native ARM64 PE machine type.
- Verified MSVC compile and link actions contain `/MT`, ThinLTO, and the declared FDO profile, with no `/MD`.
- Verified MSVC executables import Windows system DLLs only and do not import the VC or UCRT redistributable DLLs.
- Ran the packaged x64 MSVC LLVM binary under Wine and confirmed Clang 23.1.0 reports `x86_64-pc-windows-msvc`.
- Passed the focused Windows MSVC analysis and generated-action suites.
- Passed buildifier, Gazelle diff, public Starlark documentation generation, shell syntax checks, and `git diff --check`.

## Known limitations and next steps

This PR enables production of the MSVC archives but does not publish them or update the published prebuilt index, because the corresponding release assets do not exist yet.

After merge:

1. Trigger and publish the LLVM prebuilt release containing both GNU and MSVC Windows assets.
2. Add the published MSVC archive metadata and implement native Windows MSVC execution-toolchain support in a separate PR.


3rd_party/llvm-project/21.x/patches/clang-bazel-static-windows.patch
```diff
@@ -0,0 +1,36 @@
...
++    flag_values = {"@rules_cc//cc/compiler:compiler": "clang-cl"},
++    values = {"dynamic_mode": "off"},
++)
++
++config_setting(
++    name = "is_windows_msvc_static",
++    constraint_values = ["@platforms//os:windows"],
++    flag_values = {"@rules_cc//cc/compiler:compiler": "msvc-cl"},
++    values = {"dynamic_mode": "off"},
++)
```
cerisier, line 18: Is this really valid all the time ? That is, there aren't any other configuration that can yield to require CLANG_BUILD_STATIC ?

3rd_party/llvm-project/21.x/patches/llvm-bazel-blake3-windows-gnu.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/21.x/patches/llvm-bazel-windows-arm64-config.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/22.x/patches/lld-coff-lto-weak-alias-prevailing.patch
cerisier, line 1: Issue filed but fix not yet upstreamed FYI

3rd_party/llvm-project/22.x/patches/llvm-bazel-blake3-windows-gnu.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/22.x/patches/llvm-bazel-windows-arm64-config.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/23.x/patches/lld-coff-lto-weak-alias-prevailing.patch
cerisier, line 1: Issue filed but fix not yet upstreamed.

3rd_party/llvm-project/x.x/patches/llvm-bazel-blake3-windows-clang-cl.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/x.x/patches/llvm-bazel-msvc-trapping-math.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

3rd_party/llvm-project/x.x/patches/llvm-bazel-windows-arm64-config.patch
cerisier, line 1: upstreamed and merged, after llvm 23.1.0 was cut.

prebuilt/llvm/BUILD.bazel
```diff
@@ -9,6 +9,17 @@ llvm_release(
     bin_suffix = ".exe",
 )
 
+llvm_release(
+    name = "windows_msvc_llvm_release",
```
cerisier, line 13: So a single target is not sufficient now to build for all targets ?

toolchain/bootstrap/declare_toolchains.bzl
```diff
@@ -154,6 +198,35 @@ def declare_tool_map(exec_os, exec_cpu, prefix = None, fdo_profile = None, fdo_i
         strip_prefix = "clang/lib/Headers",
     )
 
+    cc_args(
+        name = prefix + "/clang_cl_compile_resource_dir",
```
cerisier, line 214: Why is this necessary here ? it feels like some semantic in the toolchain is not done right if we need an override here ? this file was supposed to be free of cc_args.

toolchain/bootstrap/declare_toolchains.bzl
```diff
@@ -417,7 +526,22 @@ def declare_toolchains(*, execs = None, targets = SUPPORTED_TARGETS):
...
+                        ":%s/clang_cl_compile_resource_dir" % tool_prefix,
+                        "@llvm//toolchain/args/windows/msvc:normalized_sdk_compile_args",
+                    ],
+                    "@llvm//platforms/config:windows_aarch64_msvc": [
+                        "@llvm//toolchain/args/windows/msvc:normalized_default_libs_for_runtime",
+                        ":%s/clang_cl_compile_resource_dir" % tool_prefix,
+                        "@llvm//toolchain/args/windows/msvc:normalized_sdk_compile_args",
+                    ],
+                    "//conditions:default": [],
+                }),
```
cerisier, line 534: Can you remind why we need to pass those here ?

toolchain/features/msvc/BUILD.bazel
```diff
@@ -48,6 +48,12 @@ _UNSUPPORTED_FEATURES = [
     "xray",
 ]
 
+_UNSUPPORTED_FEATURES_WITHOUT_FDO = [
+    feature
+    for feature in _UNSUPPORTED_FEATURES
+    if feature != "fdo_optimize"
+]
```
cerisier: Why this is needed ?

toolchain/runtimes/with_cfg_runtimes_common.bzl
```diff
@@ -12,6 +12,16 @@ def configure_builder_for_runtimes(builder, runtime_stage, linkmode = "static",
...
     builder.set("host_linkopt", [])
 
+    # Release policy is expressed through compiler-personality features rather
+    # than user compile options. Reset it at the same runtime boundary as the
+    # options above: each runtime owns its exception, RTTI, and frame-pointer
+    # requirements independently from the final LLVM product.
+    builder.extend("features", [
+        "-no_exceptions",
+        "-no_rtti",
+        "-omit_frame_pointer",
```
cerisier, line 22: This seems to have a quite fundamental change compared to before ? IIUC, we re-enable all of those for runtimes where prior to this PR, they were not ??

dzbarsky (approved):

zbarsky-openai (commented):
Adversarial review of `4c0a163`. I would fix the P1 findings before relying on these as optimized, hermetic MSVC releases. The reproduction details and the lower-priority items are included here in one batch.

### Correctness and compatibility

1. **P1 — The MSVC Stage 3 C++ binaries do not match the generated FDO profiles.** The [MSVC training rules](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/stage3/BUILD.bazel#L60-L89) run Linux-built instrumented Clang/LLVM executables, then the [MSVC feature](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/features/clang_cl/BUILD.bazel#L235-L249) applies those records to Windows/MSVC-built LLVM and suppresses missing/out-of-date warnings. The training compiler’s Windows *output* target does not change its own Linux/Itanium C++ ABI. Frontend instrumentation uses the LLVM function name as the profile key. I reproduced this: the profile recorded `_ZN11llvm_review8exerciseEi`; the Unix build received `!prof`, the Windows build’s `?exercise@llvm_review@@YAHH@Z` did not, and unmangled `main` still matched. Most LLVM/Clang/LLD C++ functions therefore cannot consume these records. Please train a compatible instrumented Windows binary (an emulator is one option) or implement a demonstrated translation, and assert nonzero profile use for named Clang/LLD C++ functions rather than checking only the command-line flag. [Clang also documents that its built-in remapping does not support the Windows C++ ABI.](https://clang.llvm.org/docs/UsersManual.html#profile-remapping)

2. **P1 — The COFF weak-alias fix still loses aliases across object files.** The [patch](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/3rd_party/llvm-project/23.x/patches/lld-coff-lto-weak-alias-prevailing.patch#L15-L35) identifies the alias’s defining bitcode file from the globally prevailing *aliasee*. If A defines COMDAT `scalar_dtor` but not `vector_dtor`, and B defines the same COMDAT plus the only `weak alias vector_dtor -> scalar_dtor`, A can win `scalar_dtor` without defining the alias, and B cannot prevail for its alias. Moreover `add(A)` can replace the fallback with `Undefined` before B’s per-file snapshot; the patch’s one-file regression cannot detect this. Track which input actually defines the selected weak alias and add a two-file regression in both link orders, including the distributed ThinLTO final link. I checked [bitcode symbol registration](https://github.com/llvm/llvm-project/blob/llvmorg-23.1.0/lld/COFF/InputFiles.cpp#L1400-L1450) and [the per-file LTO loop](https://github.com/llvm/llvm-project/blob/llvmorg-23.1.0/lld/COFF/SymbolTable.cpp#L1498-L1508); I did not have a built copy of this unpublished linker patch.

3. **P1 — Source-built `clang-cl` can execute an undeclared `lld-link` from `PATH`.** The [bootstrap declaration](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/declare_toolchains.bzl#L240-L275) copies `clang++` so that its linker is found, but leaves `clang-cl` as the default symlink to `llvm.stripped`, with `lld-link` merely staged beside the symlink. LLVM 23 resolves the real executable directory; the [MSVC `-no-canonical-prefixes` argument](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/args/BUILD.bazel#L275-L305) is frontend-only and uses `/clang:`, which is not recognized by the driver's early scan. With this project's published LLVM 23.1.0, a staged symlink and competing executable on `PATH`, both the current flags and `/clang:-no-canonical-prefixes` selected the executable on `PATH`; raw `-no-canonical-prefixes` selected the declared sibling. This can fail or run the wrong linker when execution preserves symlinks. Please copy `clang-cl`, as for `clang++`, or select the declared linker explicitly; add an execution test with a conflicting `lld-link` on `PATH`.

4. **P1 — `CLANG_BUILD_STATIC` is missing in ordinary static Bazel builds.** The [patch](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/3rd_party/llvm-project/x.x/patches/clang-bazel-static-windows.patch#L8-L33) only defines it for global `--dynamic_mode=off`. [Bazel `cc_binary` defaults to static user-library linkage](https://bazel.build/reference/be/c-cpp#cc_binary.linkstatic), and the [LLVM multicall target does not override that](https://github.com/llvm/llvm-project/blob/llvmorg-23.1.0/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel#L844-L851). Direct Windows source Stage 1/3 or archive targets do not set `dynamic_mode`, so ordinary builds without `--config=release` declare statically linked Clang components as `__declspec(dllimport)` and can fail to compile or link. Please base the macro on actual Clang DLL versus static linkage, and build a Windows Clang target under `--dynamic_mode=default`; the new provider-only test currently asserts the problematic absence.

5. **P2 — The patch changes the existing GNU archives’ default target ABI.** The [Windows config patch](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/3rd_party/llvm-project/x.x/patches/llvm-bazel-windows-arm64-config.patch#L38-L44) changes x64 GNU-built Clang from upstream `x86_64-pc-win32` (MSVC) to `x86_64-w64-windows-gnu`; it also changes the old ARM64 GNU build from the x64/MSVC fallback to native ARM64/GNU. The ARM64 correction is useful, but callers running the old archive’s `clang.exe` or `clang++.exe` without `--target` can change C++ ABI and default header/library lookup. This conflicts with the stated unchanged-GNU behavior. Please either scope this PR to MSVC or explicitly document and test the GNU default-target change as a compatibility correction.

6. **P2 — PR CI never runs the new source-stage weak-alias regression or either focused MSVC script.** The [new test](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/e2e/rules_cc/BUILD.bazel#L416-L426) requires explicit `windows/abi:msvc`; default Linux/Windows `bazel test //...` does not select it, and [CI’s explicit MSVC builds](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/.github/workflows/ci.yaml#L65-L93) only build the existing basic cross-compilation examples. The [only explicit new invocation](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/e2e/rules_cc/windows_msvc_action_test.sh#L269-L295) is in a shell script not called by CI. Please run the focused scripts or explicit source-stage regression in Linux CI and add an MSVC Stage 3/profile check; the existing 21/22 job only analyzes host Stage 1.

7. **P2 — The new MSVC archive’s named executables are POSIX symlinks.** The [new archive](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/prebuilt/llvm/BUILD.bazel#L12-L21) reuses a [manifest](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/prebuilt/llvm/llvm_release.bzl#L23-L35) that writes only `llvm.exe` as a regular executable; `clang-cl.exe`, `lld-link.exe`, `llvm-ar.exe`, etc. are symbolic links. Windows’ built-in tar uses libarchive, and [Windows symbolic-link creation requires elevation or Developer Mode for ordinary users](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw). A clean non-elevated machine can fail to unpack the entrypoints; running only `llvm.exe` under Wine does not cover this. This is inherited from the GNU archive, but also affects this new native MSVC artifact. Please use archive hardlinks or regular launchers and test extraction and the named tools on non-elevated Windows.

### Hardening and simplification

8. **P2 — Verify the release artifacts before attesting/publishing.** The [release script](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/.github/workflows/llvm-prebuilt.sh#L48-L75) only checks that each archive exists before hashing it; the workflow then attests and publishes. Turn the reported manual checks into release assertions: PE machine on both CPUs, default triples, forbidden VC/UCRT DLL imports, archive entries, and an x64 packaged `clang-cl` compile/link plus `lld-link` invocation.

9. **P2 — Train the C++ frontend once profile compatibility is fixed.** The new [workload](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/stage3/BUILD.bazel#L60-L75) uses only zstd C files and [forces `/TC`](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/fdo_profile.bzl#L44-L61); the rule also filters out non-`.c` files. Add at least one `/TP` workload with templates and Windows/MSVC C++ semantics, and assert that it contributes profile records. Separately, the [registered Stage 3 toolchain](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/declare_toolchains.bzl#L492-L509) still hard-codes the generic profile; if only the direct Windows Stage 3 target is intended to receive MSVC training, document that distinction.

10. **P3 — Reduce duplicated declarations and patches.** The new [bootstrap tool map/resource arguments](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/bootstrap/declare_toolchains.bzl#L88-L126) duplicate the [prebuilt tool declarations](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/llvm/llvm.bzl#L186-L228); the implementations already differ in `llvm_profdata` registration and builtin include-directory allowlisting. Make the intentional differences parameters of a shared declaration. The COFF weak-alias patches are also byte-for-byte identical in 21/22/23; point them at one shared patch and check patch application for each supported major. The [major-version map](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/extensions/llvm.bzl#L69-L94) also applies the 23 patch set to 24, although the [trapping-math/BLAKE3 changes](https://github.com/llvm/llvm-project/blob/main/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel) and [Windows target changes](https://github.com/llvm/llvm-project/blob/main/utils/bazel/llvm-project-overlay/llvm/config.bzl) are already in upstream main; exclude those backports from 24 before adding a 24 archive.

11. **P3 — Document and simplify the archive targets.** The [packaging documentation](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/README.md#L352-L366) still lists only the two GNU Windows targets. Add `for_windows_amd64_msvc` and `for_windows_arm64_msvc` and spell out the ABI suffixes. The [new Windows archive rule](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/prebuilt/llvm/BUILD.bazel#L7-L43) has the same archive layout as the old Windows rule; consider sharing the underlying tar target while keeping distinct ABI-specific public wrappers and asset filenames.

12. **P3 — Cover the runtime-feature reset outside MSVC.** The [shared runtime transition](https://github.com/hermeticbuild/hermetic-llvm/blob/4c0a163d582b73522ae947c8887f6364728d61ae/toolchain/runtimes/with_cfg_runtimes_common.bzl#L15-L23) appends the three feature disables without an OS/ABI condition. The new assertions check an MSVC libc++abi action, not Linux/macOS/MinGW runtime actions. Please add representative before/after action assertions, or limit the reset to the runtimes/configurations that require it, so the stated non-Windows behavior is demonstrated.

Verification scope: targeted Clang/profile reproductions and source/workflow inspection; I did not rerun the full Bazel release builds or a native Windows extraction.

-zbarskybot

PR #727, opened 2026-08-30
Add native Windows execution for MSVC targets

cerisier:
## Summary

- register the published GNU/MinGW Windows LLVM archives as Stage 0 execution toolchains for MSVC targets
- enable native Windows x64 and ARM64 execution for all source-built bootstrap stages
- preserve native `.exe` suffixes when bootstrap tools are copied into stage directories

## Contract and root cause

An MSVC target toolchain may run with GNU/MinGW-built clang-cl: the execution ABI determines whether the compiler can run, while the explicit target triple and MSVC feature configuration determine the emitted ABI.

Stage 0 already satisfies that contract with the Windows GNU prebuilt. Source bootstrap was blocked because `bootstrap_binary` copied native Windows executables into extensionless outputs. Windows subprocess spawning then failed when tools such as `llvm-nm` were invoked. Stage 3 also resolved its FDO profile relative to consuming repositories instead of `@llvm`.

This change fixes those owning layers and registers the same supported Windows execution platforms for Stages 1–3. No MSVC-built execution prebuilt is needed.


toolchain/bootstrap/bootstrap_binary.bzl
```diff
@@ -73,7 +73,13 @@ def _bootstrap_binary_impl(ctx):
     actual = ctx.attr.actual[0][DefaultInfo]
     exe = actual.files_to_run.executable
 
-    out = ctx.actions.declare_file(ctx.label.name)
+    output_name = ctx.label.name
+    if ctx.target_platform_has_constraint(
```
dzbarsky, line 77: i guess target is exec here? Might be worth a comment?

dzbarsky (approved):

PR #728, opened 2026-08-31
Add Windows MSVC module maps and layering checks

cerisier:
## Summary

Add real Bazel C++ header module-map and `layering_check` support to the Windows MSVC clang-cl target toolchains on x86-64 and ARM64.

## Changes

- Generate MSVC crosstool maps from the Clang resource headers, libc++, filtered COM and VCRuntime trees, and UCRT/shared/UM/WinRT SDK roots.
- Keep the MinGW map composition and `_mingw.h` textual-header treatment unchanged.
- Implement `module_maps`, `use_module_maps`, and `layering_check` for clang-cl with `/clang:` transport, including current and direct dependency maps, strict declaration-use, private-header diagnostics, and module-map-home behavior.
- Retain `-fno-cxx-modules`; header layering does not claim named-module support.


runtimes/module_map.bzl
```diff
@@ -80,16 +91,21 @@ def _include_path_impl(ctx):
         else:
             textual_headers_depsets.append(src[DirectoryInfo].transitive_files)
 
+    for directory in ctx.attr.umbrella_directories:
+        submodule_directory_paths.append(directory[DirectoryInfo].path)
```
dzbarsky: hmm does that break path mapping?

cerisier: Fuck let me dig

keith (approved):
args look right to me

PR #729, opened 2026-08-31
Support parse_headers for Windows MSVC targets

cerisier:
## Summary

Implements `parse_headers` for the `clang-cl` dialect.

## Root cause

The MSVC toolchain set supports_header_parsing to false, rejected parse_headers as unsupported, and mapped header parsing directly to clang-cl without the marker wrapper. Transported -x options are not sound here: clang-cl reparses /clang: operands after its CL inputs, so a transported language selection arrives after the header input.

## Change

- Advertise parse_headers for MSVC and route both packaged and Stage1 tool maps through the execution-platform marker wrapper.
- Select C++ with native /TP by default; parse_headers_as_c selects /TC and C17. Both modes use syntax-only clang-cl execution.
- Retain module-map, layering, SDK VFS, CRT, ThinLTO/FDO/link, and unsupported-feature boundaries.
- Add x86-64/ARM64 remote action, artifact, input-closure, Stage1, and fail-closed C-mode coverage plus user-facing boundary documentation.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/toolchain/cc_toolchain.bzl
+++ b/toolchain/cc_toolchain.bzl
@@ -1,7 +1,7 @@
 load("@rules_cc//cc/toolchains:feature_set.bzl", "cc_feature_set")
 load("@rules_cc//cc/toolchains:toolchain.bzl", _cc_toolchain = "cc_toolchain")
 
-_WINDOWS_MSVC_SUPPORTS_HEADER_PARSING = False
+_WINDOWS_MSVC_SUPPORTS_HEADER_PARSING = True
 
 def cc_toolchain(
         name,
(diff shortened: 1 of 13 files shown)


toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -619,6 +619,67 @@ cc_feature(
...
+        "PARSE_HEADER": "{output_file}",
+    },
+    format = {
+        "output_file": "@rules_cc//cc/toolchains/variables:output_file",
+    },
+    requires_not_none = "@rules_cc//cc/toolchains/variables:output_file",
+)
+
+cc_args(
+    name = "parse_headers_as_cpp_args",
```
dzbarsky, line 645: ooh should we do this for normal clang too? I think we had some warnings from passing cpp args to c files for header parsing? or it already works now?

toolchain/cc_toolchain.bzl
```diff
@@ -1,7 +1,7 @@
 load("@rules_cc//cc/toolchains:feature_set.bzl", "cc_feature_set")
 load("@rules_cc//cc/toolchains:toolchain.bzl", _cc_toolchain = "cc_toolchain")
 
-_WINDOWS_MSVC_SUPPORTS_HEADER_PARSING = False
+_WINDOWS_MSVC_SUPPORTS_HEADER_PARSING = True
```
dzbarsky, line 4: do we need this const or should we inline it?

toolchain/features/clang_cl/BUILD.bazel
```diff
@@ -731,6 +794,7 @@ cc_feature_set(
         ":def_file",
         ":module_map_home_cwd",
         ":module_maps",
+        ":parse_headers_wrapper",
```
keith, line 797: maybe worth a comment that this is about command line ordering

keith (approved):

PR #733, opened 2026-09-04
Add osx.from_host() for host discovery

### LINKED ISSUE #67: Support user provided macOS SDK

### LINKED ISSUE #68: Support host discovered macOS SDK

keith:
Fixes https://github.com/hermeticbuild/hermetic-llvm/issues/68
Fixes https://github.com/hermeticbuild/hermetic-llvm/issues/67 for whenever we'd want to add other platform support

Closes https://github.com/hermeticbuild/hermetic-llvm/pull/672

### DIFF
--- a/extensions/osx.bzl
+++ b/extensions/osx.bzl
@@ -12,32 +12,95 @@ _DEFAULT_FRAMEWORKS = [
     "SystemConfiguration",
 ]
 
-def _get_from_archive(mctx):
-    module_selected_archive = None
+def _get_sdk_source(mctx):
+    running_on_macos = mctx.os.name.startswith("mac os")
+    force_host_sdk = mctx.getenv("BAZEL_MACOS_USE_HOST_SDK") == "1"
+    root_source = None
+    dependency_source = None
 
     for mod in mctx.modules:
         module_archives = [tag for tag in mod.tags.from_archive]
         if len(module_archives) > 1:
             fail("Only 1 osx.from_archive(...) tag is allowed per module")
 
-        if not module_archives:
-            continue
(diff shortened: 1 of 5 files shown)


extensions/osx.bzl
```diff
@@ -12,32 +12,95 @@ _DEFAULT_FRAMEWORKS = [
...
-    fail("Missing osx.from_archive(...): set osx.from_archive(urls = [...], sha256 = ..., strip_prefix = ..., type = ...) in your MODULE.bazel")
+    fail("Missing osx.from_archive(...): osx.from_host() and BAZEL_MACOS_USE_HOST_SDK=1 only apply when Bazel is running on macOS")
+
+def _host_macos_sdk_repository_impl(rctx):
+    result = rctx.execute(
+        ["/usr/bin/xcrun", "--sdk", "macosx", "--show-sdk-path"],
+        environment = {
+            "DEVELOPER_DIR": rctx.os.environ.get("DEVELOPER_DIR", ""),
+        },
+    )
```
cerisier, line 56: Thank you for that ! Do you think we should auto discover or use a hardcoded path ? Happy with either :)

cerisier (approved):

extensions/osx.bzl
keith: i think we have to autodiscover because this could be a path to xcode, not just the CLT, and those could be installed anywhere

PR #734, opened 2026-09-05
Declare shared ThinLTO backend features

dzbarsky:
Declare `thin_lto_linkstatic_tests_use_shared_nonlto_backends` and `thin_lto_all_linkstatic_use_shared_nonlto_backends` in the generic and MSVC toolchains' known features. Both remain disabled by default. Previously Bazel silently ignored requests for these undeclared features, preventing users from selecting shared ThinLTO backends.

Add three tests against the real LLVM toolchain that verify the defaults and independent selection of each feature. All three pass on the Linux x86_64 toolchain with Bazel 9.2.0 and remote execution. A static C++ test calling a Rust library also compiles, links, and runs successfully with these declarations, fixed rules_cc, and both `thin_lto` and `thin_lto_linkstatic_tests_use_shared_nonlto_backends` enabled.

Related to https://github.com/hermeticbuild/rules_rust/issues/49. Rust PIC libraries also need rules_cc commit 4591082affa5a63f4f06942e14ea69984bb68ed0, which creates the PIC shared-backend map.

Regression coverage: https://github.com/hermeticbuild/rules_rust/pull/52 and https://github.com/bazelbuild/rules_cc/pull/868.

-zbarskybot

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,6 +65,7 @@ jobs:
           repository_cache_flags=()
           if [[ "${{ matrix.folder }}" == "e2e/cross_compilation" ]]; then
             repository_cache_flags+=(--repository_cache= --repo_contents_cache=)
+            repository_cache_flags+=(--jobs=100)
           fi
 
           bazel \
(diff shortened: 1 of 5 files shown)


cerisier (approved):
TIL

.github/workflows/ci.yaml
```diff
@@ -65,6 +65,7 @@ jobs:
           repository_cache_flags=()
           if [[ "${{ matrix.folder }}" == "e2e/cross_compilation" ]]; then
             repository_cache_flags+=(--repository_cache= --repo_contents_cache=)
+            repository_cache_flags+=(--jobs=100)
```
cerisier, line 68: is this on purpose ?

cerisier (approved):

.github/workflows/ci.yaml
dzbarsky: Not really but I think codex is trying to prevent OOM haha

PR #740, opened 2026-09-08
Add support for Clang header modules

fmeum:
First part of #676, which this splits into separately landable pieces.

Adds the `header_modules`, `use_header_modules`, `header_module_compile`, `header_module_codegen` (with `header_modules_codegen_functions` and `header_modules_codegen_debuginfo`) features to the rules-based toolchain, mirroring Google's crosstool (see [rules_android_ndk's `ndk_cc_toolchain_config.bzl`](https://github.com/bazelbuild/rules_android_ndk/blob/8ba0720dcf044fc1cdd82ca8160df0dd394f8711/ndk_cc_toolchain_config.bzl#L538)). Modules are built with `-fmodules-local-submodule-visibility`; module codegen is opt-in.

Toolchain specifics:

- The module compile args (`-xc++ -Xclang -emit-module ...`) are toolchain args gated on `header_module_compile` rather than args of the feature, since `-x` has to precede the `-c <source>` emitted by the legacy features and feature args of per-target features render after those.
- `use_header_modules` passes `-fmodule-name` and `-fmodule-map-file` itself, which `rules_cc` otherwise only passes with `layering_check`. Header modules work with and without `layering_check`, which remains recommended.

Requires Bazel 9.3.0 or later, which adds imported module files to the inputs of module codegen actions (https://github.com/bazelbuild/bazel/pull/30396); `.bazelversion` is bumped to 9.3.0rc2 accordingly.

Modular targets can't include system headers yet: the toolchain's module map covers them with umbrella submodules, which `-fmodules` builds can only include via a compiled module, and the whole sysroot can't be compiled into one (umbrella submodules make Clang parse every header beneath them, including ones that aren't standalone-parseable). #760 addresses this.

The third commit fixes the `-Wunused-command-line-argument` warnings reported below: `-fmodule-file-deps` is dropped, since rules_cc never requests a dependency file for compiles that use header modules, and module codegen actions no longer receive include paths, which they can't use since they compile a precompiled module (Bazel's legacy `include_paths`, `includes` and `preprocessor_defines` features exclude the action, rules_cc's rules-based counterparts use `source_compile_actions`, which includes it; the `include_flags` feature is inlined with the action excluded until that is fixed upstream). The tests now compile with `-Werror` so that such warnings fail the build.


toolchain/BUILD.bazel
```diff
@@ -338,6 +338,9 @@ cc_args_list(
         "//toolchain/args:no_absolute_paths_for_builtins",
         "//toolchain/args:deterministic_compile_flags",
         "//toolchain/args:module_map",
+        # Only affects cpp_module_compile actions, which are only created for
+        # targets with the `header_modules` feature enabled.
```
fmeum:
```suggestion
```

keith:
fyi looks like this warning comes up sometimes:

```
clang++: warning: argument unused during compilation: '-fmodule-file-deps' [-Wunused-command-line-argument]
```

fmeum:
Nice catch, I sent https://github.com/bazelbuild/rules_cc/pull/897 and added a workaround.

toolchain/BUILD.bazel
```diff
@@ -518,6 +520,20 @@ cc_artifact_name_pattern(
...
+# unused arguments. Bazel's legacy include_paths, includes and
+# preprocessor_defines features don't cover the action either.
+# TODO: Use @rules_cc//cc/toolchains/actions:source_compile_actions after
+# https://github.com/bazelbuild/rules_cc/pull/897.
+cc_action_type_set(
+    name = "source_compile_actions",
+    actions = ["@rules_cc//cc/toolchains/actions:source_compile_actions"],
+    excludes = ["@rules_cc//cc/toolchains/actions:cpp_module_codegen"],
+    visibility = ["//visibility:public"],
+)
```
cerisier, line 535: I'll create an issue so that we remember to revert this when we bump rules_cc.

cerisier (approved):
LGTM !

PR #741, opened 2026-09-08
Move the BuildBuddy API key from CI command lines to .bazelrc

fmeum:
`duplicate_static_library_validator_test` runs a nested Bazel with the workspace's `.bazelrc`, which imports the root `.bazelrc` and thus points the BES, remote cache and remote downloader at BuildBuddy. The API key that CI passed to the outer invocation via `--remote_header` didn't reach it, and now that BuildBuddy is phasing out anonymous access, the nested build fails in the remote downloader (seen on #739's CI run for `e2e/rules_cc` with Bazel `last_rc` on `ubuntu-24.04-arm`):

```
Error in download_and_extract: java.io.IOException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Missing API key
```

Pass the key via `.bazelrc` instead, next to the BuildBuddy endpoints it authenticates, so that every invocation picks it up, including nested and local ones, and drop the now redundant `--remote_header` flags from CI (the key was already public in the workflow file).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,7 +72,6 @@ jobs:
             --bazelrc=.bazelrc \
             test \
             "${repository_cache_flags[@]}" \
-            --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7 \
             ${{ matrix.extra_flags }} \
             //...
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -86,8 +85,7 @@ jobs:
               --bazelrc=.bazelrc \
               build \
               "${repository_cache_flags[@]}" \
-              --remote_header=x-buildbuddy-api-key=4jtaxdhxtyu4ylxdEwI7 \
-              --repo_env=BAZEL_MSVC_RUNTIME_VISUAL_STUDIO_EULA=1 \
+                --repo_env=BAZEL_MSVC_RUNTIME_VISUAL_STUDIO_EULA=1 \
```
dzbarsky: indent change?

dzbarsky (approved):
thanks

.github/workflows/ci.yaml
fmeum:
```suggestion
              --repo_env=BAZEL_MSVC_RUNTIME_VISUAL_STUDIO_EULA=1 \
```

PR #743, opened 2026-09-09
Make header_parser.c compile under MSVC

titusfortner:
`tools/internal/header_parser.c` includes `<unistd.h>` for `open`, `close` and `execv`, which MSVC doesn't ship. It is bound to `cpp_header_parsing` in the toolchain's tool map and built `[for tool]`, so on a Windows host with an MSVC exec platform every toolchain instantiation fails with `fatal error C1083: Cannot open include file: 'unistd.h'`, even with the `parse_headers` feature disabled.

The touch now uses `fopen`/`fclose`, which is portable and avoids `_open` rejecting the `0666` mode through the invalid parameter handler on Windows. The exec on Windows is `_spawnv(_P_WAIT, ...)` returning the child's status, matching `link_wrapper.c`, since `_execv` returns to the caller with the child still running and would lose clang's exit status. A `#define open _open` / `#define execv _execv` shim would have kept the diff include-only but carried both of those problems.

### DIFF
--- a/tools/internal/header_parser.c
+++ b/tools/internal/header_parser.c
@@ -1,9 +1,14 @@
 #include <errno.h>
-#include <fcntl.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+#ifdef _WIN32
+#include <process.h>
+#else
 #include <unistd.h>
+#endif
 
 int main(int argc, char **argv) {
   const char *path = getenv("PARSE_HEADER");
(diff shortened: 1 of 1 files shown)


cerisier (approved):
Thanks for this.

Just so I understand, how did you run into this ? I have a pending PR to support parse_headers on MSVC, which I believe was ignored until now ?

titusfortner:
The problem is compiling from a Windows host, not on an MSVC target. The Linux target toolchain's tool map includes header_parser, and it's built [for tool], so it's compiled by the exec platform's toolchain, which on that machine is MSVC. It fails at toolchain instantiation before any header parsing action runs, so disabling parse_headers on the target has no effect since the cc_tool is still an input of the toolchain. Anything with a Linux exec platform never sees it, since clang compiles the tool there.

cerisier:
I'm embarassed that CI didn't catch this... Let me have a look.

cerisier:
All good merged :) Thank you for catching that. We need to find a way to test that using distributed windows workers :/

PR #747, opened 2026-09-13
Factor runtime architecture and libc selectors

dzbarsky:
Select architecture before libc in runtime aliases and compose Zig target triples from independent selectors. Preserve default libc, ARM ABI, kernel availability, and explicit kernel overrides while avoiding architecture/libc cross-product selectors.


cerisier (approved):
That's great. I'll add this consideration to the bazel skill !

PR #748, opened 2026-09-13
Use AutoconfResultInfo for configure results

dzbarsky:
Add `AutoconfResultInfo` with the five fields consumed by `autoconf_hdr`: `name`, `kind`, `value`, `result`, and `defines_on_success`. Construct `AutoconfResultInfo` in `autoconf_config`, return result files directly from `declare_compile_probe` and `declare_link_probe`, and remove `policy_result`. This avoids retaining the complete check declarations in `AutoconfConfigInfo.results` while preserving header generation and the `autoconf_results` output group.

Saves around 90MB retained heap on our repo


cerisier (approved):
Nice, indeed we don't need to retain the full check.

PR #749, opened 2026-09-13
Add @llvm//runtimes:exceptions flag for libc++ and libc++abi

WillemKauf:
Adds a new build setting, --@llvm//runtimes:exceptions, which allows users to build libc++ and libc++abi without exceptions (e.g. add support for `-fno-exceptions`) when using hermetic-llvm. It is the Bazel form of LIBCXX_ENABLE_EXCEPTIONS=OFF and LIBCXXABI_ENABLE_EXCEPTIONS=OFF.

On MSVC targets, clang-cl already compiles the runtimes without exceptions (since /EH is not passed), and the libc++ MSVC copts pass none, so a refining config_setting (exceptions_disabled_msvc) maps the flag to no extra argument.

Also adds a test under tests/no_exceptions, which builds one probe twice for linux_x86_64_musl. Once with the default runtimes, and once through a transition that sets the exceptions=False flag, and checks the symbol tables with llvm-nm. The default link defines __cxa_throw (operator new's std::bad_alloc), whereas the flagged link defines none of __cxa_throw, __cxa_begin_catch or __gxx_personality_v0.


WillemKauf:
> Thanks for doing this. I need to check how it composes with other ways we deal with this. For instance I hacked some features which may need tweaking. I can finish on top of this PR as well !

Let me know how it goes

WillemKauf:
CI failure seems to be a transient network error

PR #751, opened 2026-09-15
Use complete Clang resource directories for linking

cerisier:
Let Clang discover its builtin headers in driver order, with the matching resource directory supplied as tool data. Link actions use a fully materialized resource directory containing headers, sanitizer support files, and target runtime libraries, so `rules_foreign_cc` can safely combine compile and link flags.

Because of resource dir composition, I renamed a few rules so that it matches semantics from directory/subdirectory with resource_directory and merged_resource_directory.

Builds on @keith's single-resource-dir approach in #388 (already closed) and @fmeum's header-ordering fix in #739.

Closes #739.


cerisier:
Good to merge from all 3 on various channels.

PR #760, opened 2026-09-18
Declare system headers as textual headers in the toolchain's module map

fmeum:
Second part of #676, stacked on #740.

The toolchain's module map covered source directories of system headers with umbrella submodules, which makes them non-textual module members: compile actions with the use_header_modules feature enabled can only include such headers via a compiled module, which the toolchain doesn't (and can't, see the previous commit) provide. Google's internal crosstool module map instead declares all sysroot headers as `textual header` entries: textual headers preserve layering_check semantics, but are parsed textually even in -fmodules builds and thus never require a compiled module.

Follow that model by declaring every header as textual. Source directories are opaque to Bazel, but headers_directory already enumerates their files in the DirectoryInfo it provides for include allowlisting, so the module map lists them from there. Output directories (e.g. the musl and from-source libc++ headers) are expanded to their constituent files when the module map is written, since Args expands tree artifacts at execution time. The host macOS SDK was already enumerated this way, so this also unifies its handling with that of the downloaded SDK.

The builtin resource directory contributes only its include subdirectory: the umbrella submodule over the whole directory also covered the runtime libraries and ignorelists beside the headers.

This also adds the sanitizer interface headers to the macOS module map, where they were missing, so that layering_check accepts including them.

With this, use_header_modules works without any compiled modules for system headers, matching Google's setup.


cerisier (approved):
Letting @dzbarsky taking a look as well but this is LGTM. 

The files glob is only used for module map generation generation, source directories are still used for compilation which was what my internal alarm raised when it saw glob().

kernel/extension/kernel-headers.BUILD.bazel
```diff
@@ -15,3 +16,16 @@ headers_directory(
     path = "include",
     visibility = ["//visibility:public"],
 )
+
+# The kernel headers as an enumerated list rather than a directory. Used to
```
dzbarsky: `headers_directory` does carry the DirectoryInfo provider with the list of files, I wonder if it can be reused? It combines the per-file provider with the source-directory action input

runtimes/module_map.bzl
```diff
@@ -74,11 +76,19 @@ def _include_path_impl(ctx):
...
 
     for src in ctx.attr.srcs:
-        if SourceDirectoryInfo in src or DirectoryInfo not in src:
-            # We're either a source directory or an output directory (Tree Artifact).
+        if SourceDirectoryInfo in src:
+            # Source directories are opaque even at execution time, so they
+            # can only be covered by umbrella submodules.
             submodule_directories.append(src[DefaultInfo].files)
-        else:
+        elif DirectoryInfo in src:
```
dzbarsky: hmm couldn't you do just this change to make it work with previous targets?

fmeum:
@dzbarsky Good catch, I was able to delete a bunch of code :-)

dzbarsky (approved):
Thanks this is way better ☺️

PR #761, opened 2026-09-18
Generate submodules and omit extern module declarations in module maps

fmeum:
Enables the `generate_submodules` and `module_map_without_extern_module` features by default, matching Google's crosstool, where the `module_maps` feature implies both (see [rules_android_ndk's `ndk_cc_toolchain_config.bzl`](https://github.com/bazelbuild/rules_android_ndk/blob/8ba0720dcf044fc1cdd82ca8160df0dd394f8711/ndk_cc_toolchain_config.bzl#L578-L590)). Independent of #740, but motivated by it.

**`module_map_without_extern_module`:** generated module maps no longer reference their dependencies' module maps by path via `extern module` declarations; the maps of direct dependencies are passed explicitly via the `dependent_module_map_files` variable instead, which `use_module_maps` already does. This matters for consistent behavior across execution strategies: transitive module maps aren't inputs of compile actions, so the declarations referencing them dangle in sandboxed and remote builds, but are followed (parsing the transitive closure of module maps) in unsandboxed ones, so the same build can succeed or fail depending on the strategy. For example, in a header modules build, a modular target depending on a non-modular target whose header includes a header of a modular dependency compiles fine sandboxed (the header is absorbed textually), but fails unsandboxed with "module is needed but has not been provided". With the feature, both behave the same.

**`generate_submodules`:** each header of a target is placed in its own submodule of the target's module. When the module is compiled with `-fmodules-local-submodule-visibility`, each header is thus parsed in its own scope, which keeps headers of the same module from relying on each other's declarations without including them, and is required for multi-header modules whose headers textually include shared internal headers (such as libc++). Textual module maps aren't affected.


cerisier (approved):
Based

PR #762, opened 2026-09-18
Declare textual headers in the toolchain's module map with their size

fmeum:
Stacked on #760.

Clang stats every textual header declared in a module map while parsing the map, unless the declaration carries a `size` (or `mtime`) attribute, in which case the header is only resolved lazily when a file of that size is included (`ModuleMap::addUnresolvedHeader`). With the several thousand system headers the toolchain's module map declares after #760, every compile that uses header modules pays for these stats: with sizes, a module compile drops from 0.15s to 0.12s and a serial sandboxed rebuild of a small modular benchmark (96 TUs importing 48 modules) from 44s to 30s on an M3 Max. Compiles without header modules are unaffected.

Adds `module-map-generator`, a small C tool built with the `stage1_hosted` toolchain like the other internal tools. It receives the textual headers via a params file, stats them and writes the module map with `size` attributes; the headers thereby become inputs of the module map action. `mtime` is deliberately not used since it isn't hermetic.

Since the tool's own toolchain has a module map, only the `complete` runtime stage uses the tool. The earlier stages keep writing the module map directly, which avoids a dependency cycle.

Verified on macOS with Bazel 9.3.0rc2: the generated map is identical to a manually size-annotated one, `//tests/header_modules/...` passes, and with the remaining commits of #676 on top, the libc++ module and the example build and pass as well.


cerisier (approved):
Neat !

PR #767, opened 2026-09-20
Build LLVM with libxml2 so that llvm-mt and lld-link manifest merging work

### LINKED ISSUE #765: Build the LLVM prebuilts with libxml2 so llvm-mt (and lld-link manifest merging) work
The `llvm-toolchain-minimal` prebuilts ship `llvm-mt` as a multicall entry of the `llvm` binary, but LLVM is built without `LLVM_ENABLE_LIBXML2`, so the tool is the compiled-in stub that rejects every invocation:

> $ llvm-toolchain-minimal-23.1.0-darwin-arm64/bin/llvm-mt /manifest a.manifest /out:out.manifest
> llvm-mt: error: no libxml2

(`a.manifest` being a minimal `<assembly manifestVersion="1.0"/>`.) The same build flag gates manifest merging in `lld-link` (`/MANIFESTINPUT:` with more than one input, or `/MANIFESTUAC` merging), which then falls back to trying `mt.exe`.

### Why it matters outside Bazel

Bazel-driven links never call `mt`, so this is invisible in hermetic-llvm's own use. Other build systems that consume the prebuilts do: CMake links MSVC-ABI executables through `cmake -E vs_link_exe`, which runs the linker with `/MANIFEST` and then unconditionally invokes `CMAKE_MT` to embed the intermediate manifest. With this llvm-mt every executable link fails unless the project passes `/MANIFEST:NO`, which drops the application manifest altogether (no `requestedExecutionLevel`, DPI awareness, `longPathAware` or supported-OS declarations). This came up while building a CMake toolchain on top of the prebuilts (hermetic-llvm-cmake), where `/MANIFEST:NO` is currently forced as a workaround.

### Proposed change

The dependency is already in the module graph: `libxml2` (2.15.1.bcr.1 from the BCR) is pulled in for libarchive's xar support. LLVM's Bazel overlay just doesn't wire it:

- define `LLVM_ENABLE_LIBXML2 1` in the generated llvm/Config/config.h (the overlay's `config.h.cmake` has the `#cmakedefine`, it's simply never set), and
- add `@libxml2` to the deps of the `WindowsManifest` library target (`llvm/lib/WindowsManifest`), which is what `llvm-mt` and lld's COFF driver link.

fmeum:
Fixes #765.

The `llvm-toolchain-minimal` prebuilts ship `llvm-mt` as a multicall entry, but LLVM was built without `LLVM_ENABLE_LIBXML2`, so it rejected every invocation with `llvm-mt: error: no libxml2`. The same flag gates the manifest merging in `lld-link` (`/MANIFESTINPUT:` with more than one input, `/MANIFESTUAC` merging), which then fell back to `mt.exe`.

This adds a `libxml2` wrapper to the Bazel overlay's `third-party` package in the same way as the existing `zlib` and `zstd` wrappers (as a source patch for LLVM 22 and 23, the LLVM 21 overlay predates `third-party/BUILD.bazel`), makes `LLVMWindowsManifest` depend on it and provides the BCR `libxml2` module as `@llvm_libxml2`. The wrapper can be disabled with `--@llvm-project//third-party:llvm_enable_libxml2=false`. I intend to send the overlay change upstream as well.

Note that `llvm-mt` still ignores `/inputresource:`, `/outputresource:` and `/updateresource:` (it prints `ignoring unsupported '...' option` and exits 0), so CMake's `vs_link_exe` flow still won't embed manifests with it. Only merging and validation via `/manifest` and `/out:` become functional.

### Size impact

Measured on the standalone `@llvm-project//llvm:llvm-mt` binary built with the `release` flags (`-c opt`, ThinLTO, `--strip=always`) and the stage 0 prebuilt compiler, with `--@llvm-project//third-party:llvm_enable_libxml2=false` vs. the default. The multicall `llvm` binary gains the same libxml2 code once, since `llvm-mt` and `lld` share `LLVMWindowsManifest`:

| Target | raw before | raw after | raw delta | `zstd -22 --ultra` before | after | delta |
|---|---|---|---|---|---|---|
| darwin-arm64 | 370,624 | 861,568 | +490,944 (+479 KiB) | 106,271 | 285,523 | +179,252 (+175 KiB) |
| linux-amd64-musl | 588,120 | 1,147,888 | +559,768 (+547 KiB) | 188,453 | 401,801 | +213,348 (+208 KiB) |

For reference, the published `llvm-toolchain-minimal-23.1.0-darwin-arm64.tar.zst` (release 3) is 34,590,336 bytes, so this is roughly +0.5% of the compressed archive. The delta is dominated by libxml2's parser, tree and serialization code (the multicall's ThinLTO + FDO build may shave off a bit more).

Verified that the resulting `llvm-mt` merges manifests:

```
$ llvm-mt /manifest a.manifest /manifest b.manifest /out:merged.manifest
```


cerisier (approved):
Should we upstream that to llvm ?

dzbarsky (approved):

PR #772, opened 2026-09-21
Add llvm-lib to the toolchain binaries

### LINKED ISSUE #771: Add `llvm-lib` to the toolchain binaries (a symlink to the existing multicall driver)
### Summary

The `llvm-toolchain-minimal` archives expose every tool as a symlink to the single multicall binary (`bin/llvm`, `bin/llvm.exe` on Windows). `llvm-lib`, the `lib.exe`-compatible front end of `llvm-ar`, is not among the names in `TOOLCHAIN_BINARIES` (`tools/defs.bzl`), although its code is already in the binary: `llvm-ar` dispatches on the name it is invoked with, and the same executable answers as `llvm-ranlib` and `llvm-dlltool`, which are listed.

Adding `"llvm-lib"` to `TOOLCHAIN_BINARIES` would add one more `bin/llvm-lib` (`.exe`) symlink in `prebuilt/llvm/llvm_release.bzl` and nothing else to the download.

### Why

Non-Bazel consumers of the prebuilts on the MSVC ABI have no `lib.exe`-style archiver. CMake's Windows platform rules, MSBuild and every tool that expects `lib.exe` syntax (`/out:`, `/nologo`, `/machine:`) need one; `llvm-ar` cannot be dropped in because its command line is `ar`-style.

Concrete case, the hermetic-llvm-cmake toolchain (https://github.com/Orphis/hermetic-llvm-cmake): with no `llvm-lib`, it overrides CMake's static-library rule to run `llvm-ar qcs`. CMake's stock `lib.exe` rule is run verbatim by the Ninja generator, because `lib.exe` always writes a fresh archive; `llvm-ar q` appends instead, so an incremental rebuild appended a second copy of every changed object to the existing `.lib`, and lld-link kept resolving symbols from the stale first member (fixed on our side by deleting the archive in the rule, Orphis/hermetic-llvm-cmake#10). With `llvm-lib` shipped, the override reduces to `set(CMAKE_AR llvm-lib)` and CMake's own rules apply unchanged.

Bazel users are unaffected either way: the toolchain maps archive actions to `llvm-ar` on every route, including the MSVC ABI one, and Bazel actions always write a new archive.

### Size

None. Measured on the 23.1.0 release (`llvm-23.1.0-3`):

| Platform | `bin/` content |
| --- | --- |
| darwin-arm64 | one 149 MB `llvm` binary, every tool a 4-byte symlink to it |
| windows-amd64 | one 156.5 MiB `llvm.exe`, 49 zero-byte NTFS symlinks (reparse points) to it, no hard links or launcher stubs |

`llvm-lib` would be the 50th symlink.

### Verification

On darwin-arm64, a symlink named `llvm-lib` to `bin/llvm` from `llvm-toolchain-minimal-23.1.0-darwin-arm64` accepts `lib.exe` syntax and produces an archive that `llvm-ar t` reads back:

```
ln -s <toolchain>/bin/llvm llvm-lib
./llvm-lib /nologo /out:t.lib greeter.cpp.obj
llvm-ar t t.lib
greeter.cpp.obj
```

So the driver is already built in; only the name is missing.

### Proposed change

```diff
 TOOLCHAIN_BINARIES = [
 ...
     "llvm-install-name-tool",
+    "llvm-lib",
     "llvm-libtool-darwin",
```

Optionally also expose it as `//tools:llvm-lib` (the alias list in `tools/BUILD.bazel` is generated from the same list, so this comes for free).

Orphis:
llvm-lib is the lib.exe-compatible front end of llvm-ar, selected by the name the multicall binary is invoked with, like llvm-ranlib and llvm-dlltool which are already listed. Adding the name gives the prebuilts one more symlink to bin/llvm and no other size.

Consumers outside Bazel on the MSVC ABI need a lib.exe-style archiver: CMake's Windows platform rules run it verbatim because lib.exe writes a fresh archive, while llvm-ar q appends, so a CMake toolchain using llvm-ar in that slot grew its .lib files on incremental rebuilds. Bazel keeps archiving with llvm-ar on every route and is unaffected.

Fixes #771


tools/defs.bzl
```diff
@@ -27,6 +27,7 @@ TOOLCHAIN_BINARIES = [
     "llvm-gsymutil",
     "llvm-ifs",
     "llvm-install-name-tool",
+    "llvm-lib",
```
zbarsky-openai, line 30: do we also need it in https://github.com/hermeticbuild/hermetic-llvm/blob/cba6af4d5786be13efcd2e43e8042dad33235c76/toolchain/bootstrap/transition_settings.bzl#L23 ?

Orphis: Not needed, the main binary for it is llvm-ar actually, so it's already there.

dzbarsky: It's part of bootstrap so it's technically only needed for completeness but I don't think you care, it's ok :)

